PR #9, opened 2018-06-27
Make it possible to specify the release URL as a project property.

swankjesse:
We have an internal Nexus server and otherwise this is a big drag.

Also don't require username + password unless actually uploading a release.
As-is our builds on Travis fail because it doesn't have credentials.

https://travis-ci.org/square/misk/jobs/397580749

### DIFF
--- a/src/main/kotlin/com/vanniktech/maven/publish/MavenPublishPluginExtension.kt
+++ b/src/main/kotlin/com/vanniktech/maven/publish/MavenPublishPluginExtension.kt
@@ -9,14 +9,14 @@ import org.gradle.api.Project
 open class MavenPublishPluginExtension(project: Project) {
   /**
    * The release repository url this should be published to.
-   * This defaults to the sonatypes url.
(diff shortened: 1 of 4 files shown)


src/main/kotlin/com/vanniktech/maven/publish/MavenPublishPluginExtension.kt
```diff
@@ -9,14 +9,14 @@ import org.gradle.api.Project
 open class MavenPublishPluginExtension(project: Project) {
   /**
    * The release repository url this should be published to.
-   * This defaults to the sonatypes url.
+   * This defaults to either the RELEASE_REPOSITORY_URL Gradle property, the system environment variable or the snapshot url.
```
ChristianBecker: Typo? This shouldn't default to the `snapshot` url.

vanniktech: Yes. I'll fix it.

swankjesse:
Thanks! Looking forward to the next release.

vanniktech:
@swankjesse just released 0.3.0

PR #23, opened 2018-09-25
Add the ability to specify targets and push to multiple maven repos.

budius:
The client code will simply create a `mavenRepositories{ }` and add info on the needed repos.
Example:
```
mavenRepositories {
	internalRepo = {
                releaseRepositoryUrl = MAVEN_URL_INTERNAL
                snapshotRepositoryUrl = MAVEN_URL_SNAPSHOT_INTERNAL
                repositoryUsername = MAVEN_USERNAME_INTERNAL
                repositoryPassword = MAVEN_PASSWORD_INTERNAL
	}
	betaRepo = {
                releaseRepositoryUrl = MAVEN_URL_BETA
                snapshotRepositoryUrl = MAVEN_URL_SNAPSHOT_BETA
                repositoryUsername = MAVEN_USERNAME_BETA
                repositoryPassword = MAVEN_PASSWORD_BETA
	}
}
```
and this will create the tasks `uploadInternalRepo` and `uploadBetaRepo`


gabrielittner:
I think it would be great if more of the code could be reused. The new `MavenPublishRepositoryModel` and the existing extension have exactly the same properties. The mavenDeployer configuration is the same except where they get their values from.

budius:
@gabrielittner I totally agree it's not the best reusable code. Now I'm slightly under-pressure to get moving with the project and got it done just to the point of having the minimum working.

It will take some time until I'll be able to come back to it and get a refactor to unify it all. If anyone feel like taking a stab at it before I come back, go for it.

But for a clean re-use possibly would need to change the plugin API.   
What I mean is, instead of `mavenPublish {` with the 4 parameters and then another `mavenRepositories {` with a map of the parameters, we could unify them.

My first suggestion would be like 
```
mavenPublish {
   archives {
       // those will be used in `uploadArchives` and automatically grab the default as in MavenPublishPluginExtension.kt
    }
   // others, those need to explicitly declared, we won't check `findProperty` or `enviroment`
```

but then I'll certainly wait for the maintainers to give a green light before changing.

gabrielittner:
I think it would already be possible to reuse some of the plugin code without breaking the API.

If we have an interface like the following both the existing extension and your new model could implement it and then we could have a shared method that does this part https://github.com/vanniktech/gradle-maven-publish-plugin/blob/68b120175d1e25b29174095cd83c89ef1afe7466/src/main/groovy/com/vanniktech/maven/publish/MavenPublishPlugin.groovy#L30-L43

```kotlin 
interface RepositoryModel {
  var releaseRepositoryUrl: String
  var snapshotRepositoryUrl: String
  var repositoryUsername: String?
  var repositoryPassword: String?
}
```

I'd be happy to do a follow up on this if you're too busy right now, but let's wait for what @vanniktech thinks.

vanniktech:
Sorry for answering late to this.

Yes, we should reuse as much as possible. Also, the task names should be prefixed with `uploadArchives` and not only with `upload`. 

Feel free to improve this @gabrielittner and create a PR against this one so we'll ship this feature in one go (to not break snapshot users - if there are any).

Also, the default of the new behavior should be exactly the same as it was before. So configuring this is optional and will just extend the functionality. The normal `uploadArchives` and `installArchives` should behave exactly like before.

https://github.com/vanniktech/gradle-maven-publish-plugin/pull/9 - should not break either. On a side note, can't we use this mechanism somehow smartly?

budius:
@gabrielittner I like it, and it makes sense to me. I gotta admit I had too much fun playing with gradle plugin and all the power that it brings so I'll try to squeeze an hour tomorrow.

@vanniktech prefix with `uploadArchives` sounds reasonable, can change that easy/fast enough.


Regarding this:
> the default of the new behavior should be exactly the same as it was before  

Do you mean on the `MavenPublishPluginExtension` auto picking some defaults from the `findProperty` or `enviroment`?

In terms of change, it would make simpler to code, but I'm not sure I agree with the behavior. Reason is, you have a default task `uploadArchive` that uses default parameters `RELEASE_REPOSITORY_URL`, etc, etc. And then separately there's the non default tasks with non default parameters. Seems to me like it's easy for a client application to add the duplicate parameter name if it really needs to, than having some default param in background and risk send wrong archive to wrong repo.

Regarding #9,if I understood the concern correctly, I've been using this method here [findProperty(String key, String defaultValue)](https://github.com/sensorberg-dev/gradle-scripts/blob/master/common-def.gradle#L35) and it works really nice, I could just bring it over to this project and apply throughout.

Also I'll leave a question about unifying into 1 API, or keep 2 separate. Which one?

budius:
oh, never mind, I just saw this https://github.com/vanniktech/gradle-maven-publish-plugin/pull/24 superseding my PR.

budius:
This got branched and improved on https://github.com/vanniktech/gradle-maven-publish-plugin/pull/25

budius:
Actually, roll back

budius:
updated PR with commits from @gabrielittner

vanniktech:
@budius feel free to update this PR once more and we can merge this

budius:
@vanniktech sorry I took too long. Had a few days off.
Updated.

PR #24, opened 2018-09-25
reuse MavenDeployer configuration

gabrielittner:
I've tested publishing a snapshot with both installArchives and uploadArchives. 

The only behavior change is that you could theoretically set the url for snapshots to null and skip the configuration of that. It seems like Gradle will try to publish the snapshot to the regular repository in that case (that's also what happens when using installArchives with snapshots). I don't think that this is a problem because someone would need to explicitly set that.

There are still some things that I will address in other PRs 
- combining the old and new extension
- share some of the task creation logic between the new tasks and installArchives
- the signing config doesn't account for new upload tasks https://github.com/gabrielittner/gradle-maven-publish-plugin/blob/dbb373ef5ffa1d946f0c9ee353e47586a6aa84d8/src/main/groovy/com/vanniktech/maven/publish/MavenPublishPlugin.groovy#L50-L53

cc @budius

### DIFF
--- a/src/main/groovy/com/vanniktech/maven/publish/MavenPublishPlugin.groovy
+++ b/src/main/groovy/com/vanniktech/maven/publish/MavenPublishPlugin.groovy
@@ -27,7 +27,7 @@ class MavenPublishPlugin implements Plugin<Project> {
       configureMavenDeployer(project, (Upload) project.uploadArchives, extension.defaultTarget)
 
       extraRepos.map.each { key, value ->
-        def taskName = "upload" + key.capitalize()
+        def taskName = "uploadArchives" + key.capitalize()
         project.logger.debug("Creating $taskName to upload to ${value.releaseRepositoryUrl} / ${value.snapshotRepositoryUrl}")
         project.tasks.create(taskName, Upload.class) {
 
--- a/src/main/groovy/com/vanniktech/maven/publish/MavenPublishPlugin.groovy
+++ b/src/main/groovy/com/vanniktech/maven/publish/MavenPublishPlugin.groovy
@@ -14,7 +14,7 @@ import org.gradle.plugins.signing.SigningPlugin
 
 class MavenPublishPlugin implements Plugin<Project> {
   @Override void apply(final Project p) {
-    p.extensions.create('mavenPublish', MavenPublishPluginExtension.class, p)
+    def extension = p.extensions.create('mavenPublish', MavenPublishPluginExtension.class, p)
     def extraRepos = p.extensions.create('mavenRepositories', MavenPublishRepositories.class)
 
     p.plugins.apply(MavenPlugin)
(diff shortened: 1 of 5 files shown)


src/main/kotlin/com/vanniktech/maven/publish/MavenPublishPluginExtension.kt
```diff
@@ -7,31 +7,52 @@ import org.gradle.api.Project
  * @since 0.1.0
  */
 open class MavenPublishPluginExtension(project: Project) {
+
+  @JvmField
+  internal val defaultTarget = MavenPublishTarget(
```
gabrielittner: My idea for this is to add the Map that is in MavenPublishRepositories to this extension and put `defaultTarget` and `localTarget` in there by default. Afterwards the plugin can do the configuration completely based on the Map.

src/main/groovy/com/vanniktech/maven/publish/MavenPublishPlugin.groovy
```diff
@@ -23,26 +23,8 @@ class MavenPublishPlugin implements Plugin<Project> {
...
-
-            configurePom(p, pom)
-          }
-        }
-      }
+    p.afterEvaluate { Project project ->
+      configureMavenDeployer(project, (Upload) project.uploadArchives, extension.defaultTarget)
 
       extraRepos.map.each { key, value ->
         def taskName = "upload" + key.capitalize()
```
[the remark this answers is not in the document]
gabrielittner: fixed 23624eb

src/main/kotlin/com/vanniktech/maven/publish/MavenPublishPluginExtension.kt
```diff
@@ -7,31 +7,52 @@ import org.gradle.api.Project
...
+    signing = false)
+
   /**
    * The release repository url this should be published to.
    * This defaults to either the RELEASE_REPOSITORY_URL Gradle property, the system environment variable or the sonatype url.
    * @since 0.1.0
    */
-  var releaseRepositoryUrl: String = project.findProperty("RELEASE_REPOSITORY_URL") as String? ?: System.getenv("RELEASE_REPOSITORY_URL") ?: "https://oss.sonatype.org/service/local/staging/deploy/maven2/"
+  var releaseRepositoryUrl: String
+    get() = defaultTarget.releaseRepositoryUrl
```
vanniktech: Will we need these in the future?

gabrielittner: We can decide wether we want to keep the old properties for easy configuration of the default `uploadArchives` or want to deprecate them in favor of the new API in the next pr.

vanniktech: Sounds good. Feel free to merge / move on.

gabrielittner:
I've tried Kotlin when I extracted the pom configuration and there was some Gradle/Groovy magic involved. Would be nice though.

budius:
Wow! I can see you guys took my idea and really too off with it. Awesome! Happy to be part!

There's certainly some groovy magic on that 'propertyMissing' that I don't think it would work anywhere else.

PR #25, opened 2018-09-25
Combined configuration and task creation for targets

gabrielittner:
The configuration looks like this:

```groovy
mavenPublish {
  // these 4 are equivalent to configuring `targets.upload` below
  releaseRepositoryUrl = "..." 
  snapshotRepositoryUrl = "..."
  repositoryUsername = "..."
  repositoryPassword = "..."
            
  targets {
	upload {
	  // configures the default uploadArchives
	}

	install {
	  // configures the default installArchives
	}

    otherRepo {
      // creates new upload task called `uploadArchivesOtherRepo`
    }
  }
}
```


src/main/kotlin/com/vanniktech/maven/publish/MavenPublishPluginExtension.kt
```diff
@@ -55,4 +57,27 @@ open class MavenPublishPluginExtension(project: Project) {
   var repositoryPassword: String?
     get() = defaultTarget.repositoryPassword
     set(value) { defaultTarget.repositoryPassword = value }
+
+  /**
+   * Allows to add additional [MavenPublishTargets][MavenPublishTarget] to publish to multiple repositories.
```
vanniktech: `[MavenPublishTargets][MavenPublishTarget]` isn't one enough?

src/main/kotlin/com/vanniktech/maven/publish/MavenPublishPluginExtension.kt
```diff
@@ -55,4 +57,27 @@ open class MavenPublishPluginExtension(project: Project) {
...
+   * @since 0.7.0
+   */
+  val targets: NamedDomainObjectContainer<MavenPublishTarget> =
+    project.container(MavenPublishTarget::class.java) { MavenPublishTarget(it, "") }.apply {
```
vanniktech: why do we need to specify the second parameter of `MavenPublishTarget`?

vanniktech:
Should we rename `upload` to `uploadArchives` and `install` to `installArchives` to have symmetry and matching names?

Probably also worth nuking the previous configuration parameters:

```
  releaseRepositoryUrl = "..." 
  snapshotRepositoryUrl = "..."
  repositoryUsername = "..."
  repositoryPassword = "..."
```

I doubt much people used them anyways.

src/main/kotlin/com/vanniktech/maven/publish/MavenPublishPluginExtension.kt
gabrielittner: The first one is what you see in when the markdown is rendered the second is the link to the class. I personally prefer that over writing `[MavenPublishTarget]s`

src/main/kotlin/com/vanniktech/maven/publish/MavenPublishPluginExtension.kt
gabrielittner: I didn't want to make `releaseRepositoryUrl` nullable so it's needs something when initializing it. Nullable and then throwing later during configuration is probably better though.

vanniktech:
Had a typo above that I corrected.

> should we also use the name of additional targets directly as task name?

I don't quite understand what you mean. Do you have an example?

gabrielittner:
The sample configuration above would result in these tasks `uploadArchives`, `installArchives` and `uploadArchivesOtherRepo`, because currently new targets are named `"uploadArchives" + target.name`. When we use the task names of the first two directly it would make sense to use `target.name` directly as task name.

src/main/kotlin/com/vanniktech/maven/publish/MavenPublishPluginExtension.kt
gabrielittner: fixed d519b3264d996175081532d924712d634e79fb54

gabrielittner:
I've renamed install and upload. The task name for additional targets should be fine as it is. I think we're good to go now.

gabrielittner:
Either @budius merges the `multiple-repos` branch into his to update the existing PR or we create a new one.

PR #30, opened 2018-10-08
improve when signing tasks run, consider all targets for signing

### LINKED ISSUE #29: Signing required not working as intended
When the plugin is only applied to modules the task graph will never contain `uploadArchives`, instead it will have `:<modulename>:uploadArchives`.

```
afterEvaluate
    signing {
        required {
            println("version ${!project.version.contains("SNAPSHOT")}")
            println("task ${project.gradle.taskGraph.hasTask("uploadArchives")}")
            !project.version.contains("SNAPSHOT") && project.gradle.taskGraph.hasTask("uploadArchives")
        }
    }
}
```

Slightly related: We should move the configuration of signing out of afterEvaluate in the plugin, required is already evaluated lazily and `sign project.configurations.archives` doesn't have to wait for anything. This would make it easier to override the plugin behavior in build scripts.

gabrielittner:
- signing configuration isn't part of afterEvaluate anymore which makes manually overriding it easier
- required (= task fails when signing isn't set up on the local machine) only considers if the task is a snapshot
- signing will only be run when also running an `Upload` task  (not during things like assemble) 
- the previous point will now consider all targets, before only `uploadArchives` was considered
- added info level logs to make debugging issues easier

fixes #29

### DIFF
--- a/src/main/kotlin/com/vanniktech/maven/publish/MavenPublishTarget.kt
+++ b/src/main/kotlin/com/vanniktech/maven/publish/MavenPublishTarget.kt
@@ -1,5 +1,8 @@
 package com.vanniktech.maven.publish
 
+import com.vanniktech.maven.publish.MavenPublishPluginExtension.Companion.DEFAULT_TARGET
+import com.vanniktech.maven.publish.MavenPublishPluginExtension.Companion.LOCAL_TARGET
+
 data class MavenPublishTarget(
   internal val name: String,
   /**
@@ -31,4 +34,13 @@ data class MavenPublishTarget(
    * @since 0.7.0
    */
   var signing: Boolean = true
-)
+) {
+
+  val taskName get(): String {
+    if (name == DEFAULT_TARGET || name == LOCAL_TARGET) {
+      return name
+    } else {
+      return DEFAULT_TARGET + name.capitalize()
+    }
+  }
+}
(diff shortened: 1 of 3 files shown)


gabrielittner:
One downside of the new onlyIf implementation is that you can't run signing explicitly. `./gradlew signArchives` will skip the task because none of the upload tasks are in the task graph.

src/main/kotlin/com/vanniktech/maven/publish/MavenPublishTarget.kt
```diff
@@ -31,4 +34,13 @@ data class MavenPublishTarget(
    * @since 0.7.0
    */
   var signing: Boolean = true
-)
+) {
+
+  val taskName get(): String {
```
vanniktech: Are there tests for this given that the API is public?

vanniktech:
> One downside of the new onlyIf implementation is that you can't run signing explicitly. ./gradlew signArchives will skip the task because none of the upload tasks are in the task graph.

Totally fine.

src/main/kotlin/com/vanniktech/maven/publish/MavenPublishTarget.kt
gabrielittner: I'll add them. I wanted to make it internal, but Groovy didn't like that for some reason.

gabrielittner:
Should be good now.

PR #32, opened 2018-10-14
experimental implementation of Configurer that uses maven-publish

gabrielittner:
Android libraries aren't supported right. We'd need a SoftwareComponent implementation for those which is tracked here: https://issuetracker.google.com/issues/37055147. Which is blocked by https://github.com/gradle/gradle/issues/1842. The latter is also preventing us of providing our own implementation of SoftwareComponent unless we're ok with internal APIs. Here is an example of how to do it without SoftwareComponent, but for Android adding the dependencies to the pom is more complex

This brings one behavior change for the old implementation. The general configuration inside the Configurers `init` is now done in afterEvaluate because we need to wait until we can read the new `useMavenPublish` property.

I still want tests for at least the generated poms. Going to look into that next week.

### DIFF
--- a/src/main/kotlin/com/vanniktech/maven/publish/MavenPublishConfigurer.kt
+++ b/src/main/kotlin/com/vanniktech/maven/publish/MavenPublishConfigurer.kt
@@ -1,0 +1,116 @@
+package com.vanniktech.maven.publish
+
+import com.vanniktech.maven.publish.MavenPublishPluginExtension.Companion.DEFAULT_TARGET
+import com.vanniktech.maven.publish.MavenPublishPluginExtension.Companion.LOCAL_TARGET
+import org.gradle.api.Project
+import org.gradle.api.component.SoftwareComponent
+import org.gradle.api.publish.maven.MavenPublication
+import org.gradle.api.publish.maven.plugins.MavenPublishPlugin as GradleMavenPublishPlugin
+import org.gradle.api.tasks.bundling.AbstractArchiveTask
+import org.gradle.plugins.signing.SigningPlugin
+import java.net.URI
+
+internal class MavenPublishConfigurer(private val project: Project) : Configurer {
+
+  private val publication: MavenPublication
+
+  init {
+    project.plugins.apply(GradleMavenPublishPlugin::class.java)
+    project.plugins.apply(SigningPlugin::class.java)
+
+    val publications = project.publishing.publications
+    publication = publications.create("maven", MavenPublication::class.java) { publication ->
+      val publishPom = MavenPublishPom.fromProject(project)
+
+      publication.groupId = publishPom.groupId
+      publication.artifactId = publishPom.artifactId
+      publication.version = publishPom.version
+
+      publication.pom { pom ->
+        pom.name.set(publishPom.name)
+        pom.packaging = publishPom.packaging
+        pom.description.set(publishPom.description)
+        pom.url.set(publishPom.url)
+
+        pom.scm {
+          it.url.set(publishPom.scmUrl)
+          it.connection.set(publishPom.scmConnection)
+          it.developerConnection.set(publishPom.scmDeveloperConnection)
+        }
+
+        pom.licenses { licenses ->
+          licenses.license {
+            it.name.set(publishPom.licenseName)
+            it.url.set(publishPom.licenseUrl)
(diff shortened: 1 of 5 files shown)


src/main/kotlin/com/vanniktech/maven/publish/MavenPublishConfigurer.kt
```diff
@@ -0,0 +1,116 @@
...
+  }
+
+  private fun MavenPublishTarget.repositoryUrl(version: String): URI {
+    val url = if (version.endsWith("SNAPSHOT")) {
+      snapshotRepositoryUrl ?: releaseRepositoryUrl
+    } else {
+      releaseRepositoryUrl
+    }
+    // the releaseRepositoryUrl is null checked in the plugin
+    return URI.create(@Suppress("UnsafeCallOnNullableType") url!!)
```
vanniktech: Could use requireNotNull(url) instead

src/main/kotlin/com/vanniktech/maven/publish/MavenPublishConfigurer.kt
```diff
@@ -0,0 +1,116 @@
...
+    return when (name) {
+      DEFAULT_TARGET -> "maven"
+      LOCAL_TARGET -> "local"
+      else -> name
+    }
+  }
+
+  private fun MavenPublishTarget.repositoryUrl(version: String): URI {
+    val url = if (version.endsWith("SNAPSHOT")) {
+      snapshotRepositoryUrl ?: releaseRepositoryUrl
```
vanniktech: Why do we default to release url here?

src/main/kotlin/com/vanniktech/maven/publish/MavenPublishConfigurer.kt
```diff
@@ -0,0 +1,116 @@
...
+      sign(publication)
+    }
+  }
+
+  override fun configureTarget(target: MavenPublishTarget) {
+    project.publishing.repositories.maven { repo ->
+      repo.name = target.repositoryName
+      repo.url = target.repositoryUrl(project.version.toString())
+      if (target.repositoryUsername != null) {
+        repo.credentials { it ->
```
vanniktech: It is the default and hence we can remove it

src/main/kotlin/com/vanniktech/maven/publish/Utils.kt
```diff
@@ -0,0 +1,12 @@
+package com.vanniktech.maven.publish
+
+import org.gradle.api.Project
+import org.gradle.api.publish.PublishingExtension
+import org.gradle.plugins.signing.SigningExtension
+import java.util.concurrent.Callable
+
+internal val Project.signing get() = extensions.getByType(SigningExtension::class.java)
```
vanniktech: Make all of them inline?

src/main/kotlin/com/vanniktech/maven/publish/MavenPublishConfigurer.kt
gabrielittner: To mirror the `uploadArchives` behavior. When no snapshot repository is specified it will always upload to the non snapshot repository. This is also how `installArchives` works which doesn't have a separate snapshot repository.

We could make it more explicit and require a `snapshotRepositoryUrl` for snapshots. For `installArchives` we'd then just set it to the same path.

src/main/kotlin/com/vanniktech/maven/publish/Utils.kt
gabrielittner: done fd48f06ffc1ea74009fc24d500f7828982a44ac9

src/main/kotlin/com/vanniktech/maven/publish/MavenPublishConfigurer.kt
gabrielittner: done fd48f06ffc1ea74009fc24d500f7828982a44ac9

src/main/kotlin/com/vanniktech/maven/publish/MavenPublishConfigurer.kt
gabrielittner: done fd48f06ffc1ea74009fc24d500f7828982a44ac9

PR #35, opened 2018-10-21
migrate general parts of the plugin to Kotlin

gabrielittner:
The diff of `MavenPublishPlugin.groovy` is nicer to view in IntelliJ because the big `if (project.plugins.hasPlugin("com.android.library")) { ... } else { ... }` only changed indentation.

This is the last migration step I'm confident in doing before we add tests that validate the outputs. The `setupConfigurerFor*` methods should be nicely migratable though afterwards.


src/main/kotlin/com/vanniktech/maven/publish/BaseMavenPublishPlugin.kt
```diff
@@ -0,0 +1,50 @@
...
+import org.gradle.api.tasks.Upload
+
+internal abstract class BaseMavenPublishPlugin : Plugin<Project> {
+
+  override fun apply(p: Project) {
+    val extension = p.extensions.create("mavenPublish", MavenPublishPluginExtension::class.java, p)
+
+    val pom = MavenPublishPom.fromProject(p)
+    p.group = requireNotNull(pom.groupId) { "groupId is required to be set" }
+    p.version = requireNotNull(pom.version) { "version is required to be set" }
```
gabrielittner: Those two weren't necessarily required before, but setting null here and in the pom doesn't really make sense.

Issue #21, opened 2018-09-16
Support for Dokka


saket:
Hey Niklas, will it be possible to add support for including javadoc generated using Dokka? Javadoc and sources for Kotlin files aren't included otherwise.

vanniktech:
Sure, want to take a stab at it?

saket:
That's what I initially wanted to try, but I'm quite bad at build tools. Would you be okay if I send potentially buggy patch and help me understand how things work?

digitalbuddha:
Not sure if its the proper way but here's how I did it in the past https://github.com/NYTimes/Store/search?utf8=%E2%9C%93&q=dokka&type=

vanniktech:
Don't have too much time currently but yes let's try to figure out a solution.

saket:
I'll try sending a patch this weekend.

saket:
Hey @vanniktech, just wanted to let you know that I tried my hands on this twice but failed terribly. Being new to Groovy and Gradle plugins, I found myself lost in the syntax. Based on @hzsweers's implementation in [AutoDispose](https://github.com/uber/AutoDispose/blob/master/gradle/gradle-mvn-push.gradle), here's what I could come up with so far: 

https://github.com/saket/gradle-maven-publish-plugin/blob/master/src/main/groovy/com/vanniktech/maven/publish/MavenPublishPlugin.groovy#L74

vanniktech:
Yup also believe this should be easier now. Will also try to have a look at this next week.

saket:
@gabrielittner that looks much better

hzsweers:
I'll take a crack at this

Egorand:
Any progress on this?

On a side note, kinda frustrated by the amount of support this library is getting, thinking about switching back to custom Gradle files.

Egorand:
Oh, found it, thanks. I'll review.

PR #37, opened 2018-12-05
Generate javadocs for Kotlin project with Dokka

Ilya-Gh:
Fixes https://github.com/vanniktech/gradle-maven-publish-plugin/issues/21

Since Dokka [does not support mixing with `JavaDocs`](https://github.com/Kotlin/dokka/issues/352#issuecomment-431342768) Dokka is now used to generate docs for both Kotlin and Java files on projects which uses `Kotlin`.

### DIFF
--- a/src/main/groovy/com/vanniktech/maven/publish/MavenPublishPlugin.groovy
+++ b/src/main/groovy/com/vanniktech/maven/publish/MavenPublishPlugin.groovy
@@ -50,7 +50,6 @@ class MavenPublishPlugin extends BaseMavenPublishPlugin {
         outputFormat 'javadoc'
         outputDirectory dokkaOutput
       }
-      androidJavadocsJar.dependsOn("dokka")
       androidJavadocsJar.configure {
         dependsOn "dokka"
         from dokkaOutput
--- a/src/main/groovy/com/vanniktech/maven/publish/MavenPublishPlugin.groovy
+++ b/src/main/groovy/com/vanniktech/maven/publish/MavenPublishPlugin.groovy
@@ -41,18 +41,25 @@ class MavenPublishPlugin extends BaseMavenPublishPlugin {
 
     def androidJavadocsJar = project.tasks.create("androidJavadocsJar", Jar.class) {
       classifier = 'javadoc'
-      from project.androidJavadocs.destinationDir
     }
 
     if (plugins.hasPlugin('kotlin-android')) {
+      def dokkaOutput = "${project.docsDir}/dokka"
       project.plugins.apply('org.jetbrains.dokka-android')
       project.dokka {
         outputFormat 'javadoc'
-        outputDirectory project.androidJavadocs.destinationDir.path
+        outputDirectory dokkaOutput
       }
       androidJavadocsJar.dependsOn("dokka")
+      androidJavadocsJar.configure {
+        dependsOn "dokka"
+        from dokkaOutput
+      }
     } else {
-      androidJavadocsJar.dependsOn("androidJavadocs")
+      androidJavadocsJar.configure {
+        dependsOn "androidJavadocs"
+        from project.androidJavadocs.destinationDir
+      }
     }
 
     project.tasks.create("androidSourcesJar", Jar.class) {
@@ -85,18 +92,24 @@ class MavenPublishPlugin extends BaseMavenPublishPlugin {
 
     def javadocsJar = project.tasks.create("javadocsJar", Jar.class) {
       classifier = 'javadoc'
-      from project.javadoc.destinationDir
     }
 
     if (project.plugins.hasPlugin("kotlin")) {
+      def dokkaOutput = "${project.docsDir}/dokka"
       project.plugins.apply('org.jetbrains.dokka')
       project.dokka {
         outputFormat 'javadoc'
-        outputDirectory project.javadoc.destinationDir.path
+        outputDirectory dokkaOutput
       }
-      javadocsJar.dependsOn("dokka")
+      javadocsJar.configure {
+        dependsOn "dokka"
+        from dokkaOutput
+      }
     } else {
-      javadocsJar.dependsOn("javadoc")
+      javadocsJar.configure {
+        dependsOn "javadoc"
+        from project.javadoc.destinationDir
+      }
     }
 
     configurer.addComponent(project.components.java)
--- a/src/main/groovy/com/vanniktech/maven/publish/MavenPublishPlugin.groovy
+++ b/src/main/groovy/com/vanniktech/maven/publish/MavenPublishPlugin.groovy
@@ -39,10 +39,21 @@ class MavenPublishPlugin extends BaseMavenPublishPlugin {
       options.linksOffline "https://developer.android.com/reference", "${project.android.sdkDirectory}/docs/reference"
     }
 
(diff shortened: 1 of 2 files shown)


Ilya-Gh:
@vanniktech Have you had a chance to look at the PR?

Ilya-Gh:
@gabrielittner Thanks for noting. javadoc might have overridden dokka files since the output dir was the same.

I moved dokka generated docs to a separate dir - `build/docs/dokka`.

src/main/groovy/com/vanniktech/maven/publish/MavenPublishPlugin.groovy
```diff
@@ -39,10 +39,28 @@ class MavenPublishPlugin extends BaseMavenPublishPlugin {
...
+    if (plugins.hasPlugin('kotlin-android')) {
+      def dokkaOutput = "${project.docsDir}/dokka"
+      project.plugins.apply('org.jetbrains.dokka-android')
+      project.dokka {
+        outputFormat 'javadoc'
+        outputDirectory dokkaOutput
+      }
+      androidJavadocsJar.dependsOn("dokka")
+      androidJavadocsJar.configure {
+        dependsOn "dokka"
```
Egorand: Does this duplicate the code on line 53? I think `androidJavadocsJar.dependsOn("dokka")` is equivalent to `androidJavadocsJar.configure { dependsOn "dokka" }`.

vanniktech:
I promise I'll try this out today and if it works, I'll publish a new version!

Egorand:
Sounds great, thank you!

vanniktech:
I've tried this out in https://github.com/vanniktech/RxBilling and when building via `installArchives`, I'm getting a lot of warnings from Dokka.

```
Can't find node by signature com.vanniktech.rxbilling.aidl.RxBillingAidlV3$DEFAULT_REQUEST_CODE
```

I don't know why it's complaining there. To my knowledge the Javadoc is correct and IDE is also able to parse and understand it.

```
Tag @see cannot be used in inline documentation.  It can only be used in the following types of documentation: overview, package, class/interface, constructor, field, method.
```

I'm not even using `@see` so why is it complaining?

Is this normal?

Ilya-Gh:
That seems to be a known issues - https://github.com/Kotlin/dokka/issues/184 and https://github.com/Kotlin/dokka/issues/90

Dokka seems to spamm in the output a lot. Unfortunatly, we do not have any alternative for Kotlin docs generation.

Ilya-Gh:
From dokka docs:
> html - minimalistic html format used by default
javadoc - Dokka mimic to javadoc

I did not check what’s the real difference is, but I think it would be better to continue using javadoc for java and kotlin, since it was used before for java.

> Also it seems like we are not setting sourceDirs & classpath. I'd like to understand why.

Dokka fetches both sourceDirs and classpath [from kotlin tasks](extractClasspathAndSourceRootsFromKotlinTasks), so it seems like the script in the comment double sets them.

vanniktech:
I've released 0.7.0 with the current master, will merge this and give people some chance to try it out. I'll also try it out myself with some of my libraries and after some incubating time, we can cut a new release.

vanniktech:
Thanks for the contribution!

Egorand:
Awesome!

src/main/groovy/com/vanniktech/maven/publish/MavenPublishPlugin.groovy
```diff
@@ -39,10 +39,27 @@ class MavenPublishPlugin extends BaseMavenPublishPlugin {
...
       classifier = 'javadoc'
-      from project.androidJavadocs.destinationDir
-    }.dependsOn("androidJavadocs")
+    }
+
+    if (plugins.hasPlugin('kotlin-android')) {
+      def dokkaOutput = "${project.docsDir}/dokka"
+      project.plugins.apply('org.jetbrains.dokka-android')
+      project.dokka {
+        outputFormat 'javadoc'
```
ZacSweers: Any reason this was chosen over `html`?

src/main/groovy/com/vanniktech/maven/publish/MavenPublishPlugin.groovy
```diff
@@ -39,10 +39,27 @@ class MavenPublishPlugin extends BaseMavenPublishPlugin {
...
+    def androidJavadocsJar = project.tasks.create("androidJavadocsJar", Jar.class) {
       classifier = 'javadoc'
-      from project.androidJavadocs.destinationDir
-    }.dependsOn("androidJavadocs")
+    }
+
+    if (plugins.hasPlugin('kotlin-android')) {
+      def dokkaOutput = "${project.docsDir}/dokka"
+      project.plugins.apply('org.jetbrains.dokka-android')
+      project.dokka {
```
ZacSweers: I don't think this is quite right, and as a result I don't think Android SDK refs get linked

See:
https://github.com/Kotlin/dokka/issues/169
https://github.com/Arasthel/SpannedGridLayoutManager/blob/master/spannedgridlayoutmanager/build.gradle#L5

I don't think the default `dokka` task will be of an android type, or at least I'm not seeing these link

src/main/groovy/com/vanniktech/maven/publish/MavenPublishPlugin.groovy
vanniktech: I was also confused - https://github.com/vanniktech/gradle-maven-publish-plugin/pull/37#issuecomment-454016415

src/main/groovy/com/vanniktech/maven/publish/MavenPublishPlugin.groovy
vanniktech: Could very well be the case. I've used the latest snapshot for publishing one library of mine but there were no Android related things in there. Mind submitting a PR?

Ilya-Gh: @ZacSweers, dokka-android plugin [creates DokkaAndroidTask](https://github.com/Kotlin/dokka/blob/master/runners/android-gradle-plugin/src/main/kotlin/mainAndroid.kt#L12) with `dokka` name when it's applied.

 The guy from the issue Kotlin/dokka#169 was using `DokkaTask` instead of `DokkaAndroidTask`.


Android SDK are not added because its missing from it's missing from configuration as [it was done for javadoc](https://github.com/vanniktech/gradle-maven-publish-plugin/pull/37/files#diff-406fe0b4199eaa97bca99d8d9ffd96d0R39). Android SDK can be linked using [externalDocumentationLink](https://github.com/Kotlin/dokka/blob/c189c8c338c219b8d773666475f5ae2818461b0a/README.md#L118) property.

I will submit a PR tomorrow with the apropriare changes if none will do it before.

keeferrourke:
We're publishing artifacts to an internal maven repo, and this dependency on `dokka` has caused some pain points. 

For anyone looking for a workaround, applying the `org.jetbrains.dokka` plugin in each of your subprojects (this will explicitly define the task before the `gradle-maven-publish` plugin does so for you) and adding the following block to your top-level `build.gradle` will skip the doc generation:

```gradle
subprojects {
 afterEvaluate { project ->
   if (project.tasks.findByName('dokka')) {
      project.tasks.dokka { onlyIf { false } }
    }
  }
}
```

Issue #40, opened 2019-01-14
Task uploadArchives looping in Kotlin artifacts.


WellingtonCosta:
When I try to upload a multi module project that has a Kotlin module, the `uploadArchives` task enter in loop and I need to force stop it.

 Any idea?

vanniktech:
Are you using the Daemon or the parallel flag? Usually, I disable those while uploading.

WellingtonCosta:
I'm using parallel flag... Would be it?

vanniktech:
Yeah. Try not using it. It should work. Might be worth pointing that out in the README. Want to submit a PR?

WellingtonCosta:
Ok. I'll try to do that without the parallel flag. If it works, I'll  submit a PR describing this issue in README.

WellingtonCosta:
It worked:

```
org.gradle.daemon=false
org.gradle.parallel=false
```

I'm sending a PR describing this in README.md.

vanniktech:
Fixed by #41

PR #41, opened 2019-01-14
Update README.md

WellingtonCosta:
Updating README.md to describe how to prevent loop behavior on Kotlin projects / modules. ( #40 )


WellingtonCosta:
Ok. I'll change this.

PR #45, opened 2019-02-02
Change docs format for Kotlin project docs

Ilya-Gh:
Initial Dokka integration implemented in https://github.com/vanniktech/gradle-maven-publish-plugin/pull/37 missed external documentation linking which resulted in Android SDK not being linked with project docs. 

I was not able to make Dokka link Android SDK using `javadocs` format, which should happen by default since Dokka [links Android SDK](https://github.com/Kotlin/dokka/blob/fc3b668e2acefad7990b8ed02b9382c0c800a04e/runners/android-gradle-plugin/src/main/kotlin/mainAndroid.kt#L36) from the box if `dokka-android` plugin is used. 

However, Dokka does link Android SDK for `Dokka` docs format.

### DIFF
--- a/src/main/groovy/com/vanniktech/maven/publish/MavenPublishPlugin.groovy
+++ b/src/main/groovy/com/vanniktech/maven/publish/MavenPublishPlugin.groovy
@@ -47,7 +47,7 @@ class MavenPublishPlugin extends BaseMavenPublishPlugin {
       def dokkaOutput = "${project.docsDir}/dokka"
       project.plugins.apply('org.jetbrains.dokka-android')
       project.dokka {
-        outputFormat 'javadoc'
+        outputFormat 'html'
         outputDirectory dokkaOutput
       }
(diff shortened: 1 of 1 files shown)


ZacSweers:
I still don't see android SDK or java SDKs linked with this change. Demo PR: https://github.com/InflationX/ViewPump/pull/27

image

Ilya-Gh:
@ZacSweers @vanniktech Thanks for the review. I changed the Dokka output format which seems to fix Android SDK linking. Can you please take another look at the changes.

vanniktech:
I'll wait for @ZacSweers

ZacSweers:
Sorry for the delay, didn't see the update! Tried again and looks like the android sdk and kotlin linking does indeed work now! I think this is a good start, though I think it'd be worthwhile to add instructions for/expose and API to add more as needed (say, the androidx libraries).

Issue #47, opened 2019-03-01
Kotlin files are not included to sources jar


lukaville:
[Here](https://github.com/vanniktech/gradle-maven-publish-plugin/blob/master/src/main/groovy/com/vanniktech/maven/publish/MavenPublishPlugin.groovy#L69) we configure androidSourcesJar with `project.android.sourceSets.main.java.sourceFiles` which doesn't contain kotlin source code. So resulting sources.jar includes only java files for kotlin projects.

Probably we should replace it with `project.android.sourceSets.main.java.srcDirs`?

vanniktech:
Probably. This can also fix the issue pointed out by @ZacSweers

vanniktech:
@lukaville let's make this change. Do you want to submit a PR or should I?

PR #48, opened 2019-03-05
Use srcDirs instead of sourceFiles to include Kotlin files to sources jar

lukaville:
#47

### DIFF
--- a/src/main/groovy/com/vanniktech/maven/publish/MavenPublishPlugin.groovy
+++ b/src/main/groovy/com/vanniktech/maven/publish/MavenPublishPlugin.groovy
@@ -66,7 +66,7 @@ class MavenPublishPlugin extends BaseMavenPublishPlugin {
 
     project.tasks.register("androidSourcesJar", Jar.class) {
       classifier = 'sources'
-      from project.android.sourceSets.main.java.sourceFiles
+      from project.android.sourceSets.main.java.srcDirs
       configurer.addTaskOutput(it)
     }
 


vanniktech:
Thank you :)

ZacSweers:
I've tried this out locally but am still not seeing kotlin sources included. Test project https://github.com/InflationX/ViewPump

vanniktech:
So, I've tested this out on https://github.com/vanniktech/RxBilling/tree/kotlin-sources

and when using `installArchives` I can see `~/.m2/repository/com/vanniktech/rxbilling-testing/0.3.0-SNAPSHOT/rxbilling-testing-0.3.0-20190311.102147-1-sources.jar` being generated. Also when I unzip that jar file I can see the Kotlin file. This is a kotlin only module.

For Android + Kotlin I've also got - https://github.com/vanniktech/Emoji/tree/kotlin-sources - which seems to be working.

I've checked out your use case and there are no sources being generated - which is weird. I've also tried around but I simply could not get it to work. It also seems weird since it does not differ a lot from the projects where I'm running it on.

I can also publish this plugin using the version from master and then I'll see sources too. Do you have any clue that can cause the fact that ViewPump does not work? @ZacSweers

vanniktech:
@ZacSweers I've created https://github.com/vanniktech/ViewPump/tree/sources which uses a local version of the plugin (you can publish it via `installArchives`) and I've noticed that the `androidJavadocsJar` task isn't called when I run `installArchives`. In my projects it is which is weird. If that one isn't called we don't do anything related with dokka so the set up that we're having is brittle.

ZacSweers:
in the end - the viewpump issue turned out to be that it was using the wrong dokka plugin artifact 😭

PR #49, opened 2019-03-14
Real integration tests

### LINKED ISSUE #42: Proper integration tests
Raised by @Egorand and a valid point. I believe we can use the Gradle Testing Kit and publish the artifacts into a local path and do actual file verification that the POM and all other content is correctly generated.

vgaidarji:
Closes #42.

### What has been done
Configured integration tests (they are called functional on official Gradle site btw).

Tests added:
- [x] Java 
- [x] Java library
- [x] Java library + Groovy
- [ ] ~Android library~ - _will be added in_ #50
- [ ] ~Android library + Kotlin Android~ - _will be added in_ #50
- [ ] ~Java library + Kotlin~ - _will be added in_ #50

 
**TODO:**
- [x] Rewrite integration test from Groovy to Kotlin
- [x] Parametrize tests to make use of both `installArchives` and `uploadArchives` tasks
- [x] Open issue to add remaining tests listed in "Tests added" section above -> #50

### Integration tests execution time
By introducing integration tests, overall **CI build time arose by ~1 min**.
Compare build from this PR ([this one for example](https://travis-ci.org/vanniktech/gradle-maven-publish-plugin/builds/506753178)) with build from `master` branch ([this one for example](https://travis-ci.org/vanniktech/gradle-maven-publish-plugin/builds/502039095)).

### How to test

Add this dirty helper function somewhere after `installArchives`/`uploadArchives` command execution in test and execute tests via `./gradlew clean integrationTest --info` command to see generated artifacts in console.
```
  fun lsGeneratedFiles() {
    val p = Runtime.getRuntime().exec("ls -R ${repoFolder.absolutePath}/com/example/test-artifact")
    Scanner(p.inputStream).use { while (it.hasNextLine()) println(it.nextLine()) }
  }
```

NO!


vgaidarji:
@vanniktech @Egorand hey there. Opening draft to share the direction and see if you have any suggestions.
The total number of tests will be the same as in `MavenPublishPluginTest` class once finished.

Egorand:
Thanks, this looks awesome!

A few thoughts:
- In order to keep PRs small and simple, I'd prefer this PR to set up a good testing framework and have just enough examples for contributors to follow. Additional tests should come in follow-up PRs. The way it looks right now is great.
- An idea for `installArchives` vs `uploadArchives`: set up a parameterized test that will run all fixtures using both tasks.
- I feel like the experience of writing tests will be better with Kotlin or Java. Unless you really need Groovy to make it work, my personal suggestion would be to switch to Kotlin.

vanniktech:
I 100% agree to what @Egorand said!

gabrielittner:
> An idea for installArchives vs uploadArchives: set up a parameterized test that will run all fixtures using both tasks.

Just to add too this: Those two are just the two default targets, so the tests should probably also support custom targets.

vgaidarji:
> In order to keep PRs small and simple, I'd prefer this PR to set up a good testing framework and have just enough examples for contributors to follow. Additional tests should come in follow-up PRs. The way it looks right now is great.

Agree, then I'll add no more tests in this PR and follow up will introduce remaining (I'll open an issue with tests to be added).

> An idea for installArchives vs uploadArchives: set up a parameterized test that will run all fixtures using both tasks.

Sounds like a plan 👍 (I'll show at the end how much time it takes to run all those tests).

> I feel like the experience of writing tests will be better with Kotlin or Java. Unless you really need Groovy to make it work, my personal suggestion would be to switch to Kotlin.

No strong preference here, I don't mind rewriting it to Kotlin.

> Just to add too this: Those two are just the two default targets, so the tests should probably also support custom targets.

Let me introduce 2 default targets first and we'll see if custom targets are required 👍

vanniktech:
Sounds great. Thanks for taking care of this!

vanniktech:
This is ready for a review now, correct?

vgaidarji:
@vanniktech yes, I've addressed all issues/suggestions except custom target in tests.

build.gradle
```diff
@@ -87,6 +88,10 @@ dependencies {
   testImplementation 'com.github.stefanbirkner:system-rules:1.18.0'
   testImplementation "org.jetbrains.kotlin:kotlin-reflect:$kotlinVersion"
   testImplementation "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlinVersion"
+
+  testCompileOnly gradleTestKit()
```
vanniktech: Why do we need this? Should not this maybe be `integrationTestCompileOnly`?

gradle/integration-test.gradle
```diff
@@ -0,0 +1,27 @@
...
+    runtimeClasspath += output + compileClasspath
+  }
+}
+
+configurations {
+  integrationTestCompile.extendsFrom testCompile
+  integrationTestRuntime.extendsFrom testRuntime
+}
+
+gradlePlugin {
```
vanniktech: What is this one referring to?

build.gradle
vgaidarji: This is required to have [`GradleTestRunner`](https://github.com/vanniktech/gradle-maven-publish-plugin/pull/49/files#diff-b9585b580f4e536ac258791d99aa04ceR4) available in tests to execute various Gradle commands in tests.
I tried initially the `integrationTestCompileOnly` too but for some unknown to me reason classes Gradle test kit were not available in integration test I've created 👀.

gradle/integration-test.gradle
vgaidarji: This is from the official [Gradle test kit docs](https://docs.gradle.org/current/userguide/test_kit.html#sub:test-kit-automatic-classpath-injection).

> The Java Gradle Plugin development plugin can be used to assist in the development of Gradle plugins. Starting with Gradle version 2.13, the plugin provides a direct integration with TestKit. **When applied to a project, the plugin automatically adds the gradleTestKit() dependency to the test compile configuration.** Furthermore, it automatically generates the classpath for the code under test and injects it via GradleRunner.withPluginClasspath() for any GradleRunner instance created by the user. It’s important to note that the mechanism currently only works if the plugin under test is applied using the plugins DSL. If the target Gradle version is prior to 2.8, automatic plugin classpath injection is not performed.

I need to double check though if I can avoid `gradleTestKit()` dependency since this plugin configuration was added here. 👍

vgaidarji: Thanks for raising this question, this construct is used to `reconfigure the conventions provided by the Java Gradle Plugin Development plugin for a project that uses a custom Test source set` which is not the case for this project and not needed.

build.gradle
vgaidarji: @vanniktech ok, it works (obviously). I can't recall already why `import org.gradle.testkit.runner.GradleRunner` wasn't working for custom integration test source set and I had to change from `integrationTestCompileOnly` to `testCompileOnly`.

build.gradle
```diff
@@ -87,6 +88,10 @@ dependencies {
   testImplementation 'com.github.stefanbirkner:system-rules:1.18.0'
   testImplementation "org.jetbrains.kotlin:kotlin-reflect:$kotlinVersion"
   testImplementation "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlinVersion"
+
+  integrationTestImplementation gradleTestKit()
+  integrationTestImplementation 'junit:junit:4.12'
```
vgaidarji: @vanniktech should we extract common dependencies versions (`junit` and `assertj`) into constants or local variables inside `dependencies` block?

vanniktech: Yes just do it similar to `$kotlinVersion` so `junitVersion = "4.12"` & `assertj3Version = "3.12.1"`

vgaidarji:
@vanniktech done 💯

vanniktech:
Thanks for your contribution. I really appreciate it!

vgaidarji:
@vanniktech thanks, always welcome! 🎉

Issue #54, opened 2019-04-12
Snapshot not uploading sources and javadocs


mzestars:
Hi, I've used the 0.8.0 version of the library and the sources and javadocs was uploaded but there is no kotlin source code in the sources.jar.

Then, I've tested the snapshot version 0.9.0-SNAPSHOT, the kotlin source code is included in the sources.jar but this time the sources and javadocs jars are not being uploaded to the repo.

Maybe there's some code you've deleted while fixing the kotlin source code issue that caused the jars not being uploaded?

mzestars:
Ok, I just looked at the source code and found out that the sources and javadocs upload task is only added if the androidSourcesJar and androidJavadocsJar is run explicitly. Thus the command is:

`./gradlew androidSourcesJar androidJavadocsJar uploadArchives --no-daemon --no-parallel`

instead of

`./gradlew uploadArchives --no-daemon --no-parallel`

vanniktech:
Huh maybe we should always run them if you execute `uploadArchives`?

mzestars:
I think having the option of not uploading them is good. Should stick with:

`./gradlew androidSourcesJar androidJavadocsJar uploadArchives --no-daemon --no-parallel`

Just that maybe u can mention this when 0.9.0 goes live. Thank u so much for an awesome library. :)

vanniktech:
Interesting. Do you get sources and javadocs when running version 0.8.0 with `./gradlew uploadArchives --no-daemon --no-parallel`?

martinbonnin:
I can reproduce the same behavior on a pure kotlin project (no android). I can include the sources + javadoc on 0.9.0-SNAPSHOT by doing `./gradlew kotlinSourcesJar javadocsJar uploadArchives --no-daemon --no-parallel`.

Running `./gradlew uploadArchives --no-daemon --no-parallel` with version 0.8.0 also includes sources + javadoc

vanniktech:
My hunch is that the task configuration avoidance in #46 introduced this.

martinbonnin:
I'd still vote to include sources and javadoc by default and not having to opt-in on the command line. Since Sonatype OSS is the default repository and they enforce sources and javadoc, it would make sens for the defaults to be consistent

vanniktech:
Yes. Definitely, a regression that needs to be fixed before we publish 0.9.0

bendb-instacart:
[...]

PR #70, opened 2019-10-08
Support for Dokka 0.10.0

### LINKED ISSUE #69: Not compatible with Dokka 0.10.0
[Dokka 0.10.0](https://github.com/Kotlin/dokka/releases/tag/0.10.0) merges the android plugin into the main plugin, so this plugin fails when it tries to apply the `dokka-android` plugin.

Error message looks like this:
```
FAILURE: Build failed with an exception.

* What went wrong:
A problem occurred configuring project ':my_project'.
> Failed to apply plugin [id 'org.jetbrains.dokka-android']
   > Cannot add task 'dokka' as a task with that name already exists.
```

zach-klippenstein:
[Dokka 0.10.0](https://github.com/Kotlin/dokka/releases/tag/0.10.0) merges the android plugin into the main plugin, so this plugin fails when it tries to apply the `dokka-android` plugin.

This change will cause the regular dokka plugin to be applied to `kotlin-android` projects instead of the obsolete `dokka-android` one, which means it **breaks support for dokka versions pre-0.10.0**.

This fixes #69.

### DIFF
--- a/src/main/groovy/com/vanniktech/maven/publish/MavenPublishPlugin.groovy
+++ b/src/main/groovy/com/vanniktech/maven/publish/MavenPublishPlugin.groovy
@@ -48,7 +48,7 @@ class MavenPublishPlugin extends BaseMavenPublishPlugin {
 
     if (plugins.hasPlugin('kotlin-android')) {
       def dokkaOutput = "${project.docsDir}/dokka"
-      project.plugins.apply('org.jetbrains.dokka-android')
+      project.plugins.apply('org.jetbrains.dokka')
       project.dokka {
         outputFormat 'html'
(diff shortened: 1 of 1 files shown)


zach-klippenstein:
@vanniktech  Confirmed that `0.9.0-SNAPSHOT` works with Dokka 0.10.0.

vanniktech:
Do you see the following issue? https://github.com/vanniktech/gradle-maven-publish-plugin/issues/54

This is what stopped me from creating a new release.

zach-klippenstein:
I'm not in a hurry to upgrade to dokka 0.10.0, so no time pressure on the release from me.

PR #80, opened 2020-01-19
More integration tests

gabrielittner:
- integration tests for Kotlin project (main addition here is the assertion that the dokka `task` runs)
- check actual pom content
- introduce `fixtures/common` which will be copied for every tests but can be overridden by the test specific fixtures
- move build files to the fixtures because they are diverging more


gradle.properties
```diff
@@ -8,13 +8,13 @@ POM_PACKAGING=jar
...
-POM_URL=http://github.com/vanniktech/gradle-maven-publish-plugin/
-POM_SCM_URL=http://github.com/vanniktech/gradle-maven-publish-plugin/
+POM_URL=https://github.com/vanniktech/gradle-maven-publish-plugin/
+POM_SCM_URL=https://github.com/vanniktech/gradle-maven-publish-plugin/
 POM_SCM_CONNECTION=scm:git:git://github.com/vanniktech/gradle-maven-publish-plugin.git
 POM_SCM_DEV_CONNECTION=scm:git:ssh://git@github.com/vanniktech/gradle-maven-publish-plugin.git
 
 POM_LICENCE_NAME=The Apache Software License, Version 2.0
-POM_LICENCE_URL=http://www.apache.org/licenses/LICENSE-2.0.txt
+POM_LICENCE_URL=https://www.apache.org/licenses/LICENSE-2.0.txt
```
vanniktech: Good one!

src/integrationTest/kotlin/com/vanniktech/maven/publish/MavenPublishPluginIntegrationTest.kt
```diff
@@ -38,48 +41,20 @@ class MavenPublishPluginIntegrationTest(
...
-            uploadArchives {
-              releaseRepositoryUrl = "file://${repoFolder.absolutePath}"
-              signing = true
-            }
-          }
-        }
-        """)
 
-    testProjectDir.newFile("gradle.properties").writeText("""
+    File("$FIXTURES/common").listFiles()?.forEach { it.copyRecursively(File(testProjectDir.root, it.name)) }
```
vanniktech: super nit: we could `testProjectDir.root.resolve(it.name)` feel free to do in another PR

src/integrationTest/kotlin/com/vanniktech/maven/publish/MavenPublishPluginIntegrationTest.kt
```diff
@@ -38,48 +41,20 @@ class MavenPublishPluginIntegrationTest(
...
+    File(testProjectDir.root, "gradle.properties").appendText("""
```
vanniktech: same here

src/integrationTest/kotlin/com/vanniktech/maven/publish/MavenPublishPluginIntegrationTest.kt
gabrielittner: I didn't know about resolve, nice 
43031c2c35292e7f1b67dbba167411acd9f8e0d1

src/integrationTest/kotlin/com/vanniktech/maven/publish/MavenPublishPluginIntegrationTest.kt
gabrielittner: 43031c2

PR #105, opened 2020-02-01
initial Kotlin Multiplatform support

### LINKED ISSUE #57: Support publishing of Kotlin Multiplatform projects
Might use SQLDelight's [gradle-mvn-mpp-push.gradle](https://github.com/square/sqldelight/blob/master/gradle/gradle-mvn-mpp-push.gradle) as a prototype. Can try to find time to implement this.

gabrielittner:
This adds support to publish multiplatform projects to all configured targets and sets up the coordinates and pom according to our properties. 

Also updated Dokka to 0.10.0 which adds experimental multiplatform support. Without that projects would have to configure some things manually to avoid all builds failing.

closes #57


build.gradle
```diff
@@ -79,8 +79,8 @@ dependencies {
   api localGroovy()
   api "org.jetbrains.kotlin:kotlin-stdlib:$kotlinVersion"
 
-  implementation "org.jetbrains.dokka:dokka-android-gradle-plugin:$dokkaVersion"
-  implementation "org.jetbrains.dokka:dokka-gradle-plugin:$dokkaVersion"
+  implementation "org.jetbrains.dokka:dokka-gradle-plugin:0.10.0"
```
vanniktech: Why did we get rid of the `dokkaVersion` variable?

gabrielittner: Wasn't intentional.

gabrielittner:
I forgot to mention that I've tested this with a local snapshot on SqlDelight. https://github.com/cashapp/sqldelight/compare/master...gabrielittner:publish-plugin
One thing that needs to happen for everything to work is updating Dokka in SqlDelight (my local snapshot was still using 0.9.x). Running the branch above with this pr would cause dokka to fail.

build.gradle
gabrielittner: fb997e6

vanniktech:
Oh wow. That's even neater.

src/main/kotlin/com/vanniktech/maven/publish/Configurer.kt
```diff
@@ -7,6 +7,8 @@ internal interface Configurer {
    */
   fun configureTarget(target: MavenPublishTarget)
 
+  fun configureMultiplatformProject()
```
Egorand: nit: `Multiplatform` -> `KotlinMultiplatform`?

gabrielittner: Makes sense https://github.com/vanniktech/gradle-maven-publish-plugin/pull/112

PR #110, opened 2020-02-04
downgrade dokka

gabrielittner:
I had troubles setting up Dokka 0.10.0 in SqlDelight in an equivalent way. Since they have a working way to use Dokka 0.9.x I don't want to force the upgrade now. 

The consequences of this PR for multiplatform projects are:
- they need to apply dokka themselves
- if they decide to use 0.9.x they need to apply the same workarounds as SqlDelight or our integration tests in here
- if they decide to use dokka 0.10.0 they need to explicitly include that version and configure it

I think long term it might even be better not to apply Dokka automatically. It's still getting used when it's there.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -2,7 +2,7 @@ buildscript {
   ext.kotlinVersion = '1.3.31'
   ext.junitVersion = '4.13-beta-3'
   ext.assertjVersion = '3.12.2'
-  ext.dokkaVersion = '0.10.0'
+  ext.dokkaVersion = '0.9.18'
 
   repositories {
     mavenCentral()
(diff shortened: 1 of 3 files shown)


ZacSweers:
> I think long term it might even be better not to apply Dokka automatically. It's still getting used when it's there.

I would be in favor of this. Inability to properly configure this plugin is part of why I haven't used it in some projects, such as ones using mkdocs (which require a different output format(

gabrielittner:
@ZacSweers The configuration issue is already solved for 0.9.0. We never set the output format and we only set the output directory when it's still empty.

Issue #82, opened 2020-01-19
Pom packaging is  not written


gabrielittner:
It's only written for Android libraries in legacy mode. I guess that AGP adds it there. The one set through this plugin never makes it to the pom file. 

Example https://repo.maven.apache.org/maven2/com/vanniktech/gradle-maven-publish-plugin/0.8.0/gradle-maven-publish-plugin-0.8.0.pom

ZacSweers:
This prevents us from using the non-legacy mode in 0.9.0 as well on https://github.com/slackhq/keeper

gabrielittner:
@ZacSweers For non Android projects it's also broken in non legacy mode. From a very cursory check it didn't even work with the Chris Bane's script 4 years ago. https://repo.maven.apache.org/maven2/com/slack/keeper/keeper/0.2.0/keeper-0.2.0.pom

gabrielittner:
Did you see any issues from not having it?

ZacSweers:
yeah our upload fails pom validation on sonatype when we tried the new API. Memory is fuzzy but it was complaining about missing POM_URL and a number of other elements

ZacSweers:
CC @jschear in case he remembers any other tidbits

gabrielittner:
Ok, will take a look. I've got a theory what it might be.

ZacSweers:
We'll try to do another repro and get a screenshot of the sonatype-reported errors

gabrielittner:
It's fine. I've just tried it out locally and can see what's going wrong. Looks like we need some special configuration for gradle plugins.

gabrielittner:
Closing this as packaging should come from the publication/component. For example Kotlin MPP sets the packaging for Kotlin/Native artifacts to `klib`.

PR #115, opened 2020-02-13
directly support java-gradle-plugin projects

### LINKED ISSUE #111: Add integration test for publishing gradle plugin marker
Currently it's just working without us having to do anything to enable it, we should a test to make sure it stays that way.

gabrielittner:
Addresses issues mentioned in #82. Like Kotlin mpp the `java-gradle-plugin` creates it's own publications. We need to configure those instead of creating our own.

Also closes #111


gabrielittner:
@ZacSweers this should fix your issues

PR #119, opened 2020-02-23
fix plugin marker pom not containing name and description

### LINKED ISSUE #117: Gradle plugin publishing: POM does not contain name and description
Needed to satisfy the maven central sync requirement. I believe this is a Gradle bug (could reproduce it with plain maven publish), still opening an issue here to track it. 
https://github.com/gradle/gradle/issues/12259

gabrielittner:
I found a workaround. Closes #117

### DIFF
--- a/src/main/kotlin/com/vanniktech/maven/publish/MavenPublishConfigurer.kt
+++ b/src/main/kotlin/com/vanniktech/maven/publish/MavenPublishConfigurer.kt
@@ -150,8 +150,12 @@ internal class MavenPublishConfigurer(
           configurePom(it, groupId = it.groupId, artifactId = it.artifactId)
           // workaround for https://github.com/gradle/gradle/issues/12259
           it.pom.withXml { pom ->
-            pom.asNode().appendNode("name", publishPom.name)
-            pom.asNode().appendNode("description", publishPom.description)
+            if (pom.asNode().get("name") == null) {
+              pom.asNode().appendNode("name", publishPom.name)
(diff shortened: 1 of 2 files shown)


src/main/kotlin/com/vanniktech/maven/publish/MavenPublishConfigurer.kt
```diff
@@ -148,6 +148,12 @@ internal class MavenPublishConfigurer(
         if (it.name == "${plugin.name}PluginMarkerMaven") {
           // keep the current group and artifact ids, they are based on the gradle plugin id
           configurePom(it, groupId = it.groupId, artifactId = it.artifactId)
+          // workaround for https://github.com/gradle/gradle/issues/12259
+          it.pom.withXml { pom ->
+            pom.asNode().appendNode("name", publishPom.name)
```
vanniktech: Is there something like `replace` or `set`. That way if Gradle were to fix their bug we won't be having two `<name>` tags?

gabrielittner: 👍 I've added checks around these. Tested it manually by adding 
```kotlin
it.pom.withXml { pom ->
  pom.asNode().appendNode("name", "1")
  pom.asNode().appendNode("description", "2")
}
```
before which resulted in the tests failing because the pom contained 1 and 2.

vanniktech:
Awesome, thanks!

PR #123, opened 2020-03-21
allow to override group and version in build files

### LINKED ISSUE #121: Version always comes from VERSION_NAME
If at configure time you update `project.version` to be something like `"${VERSION_NAME}-${buildnumberorsomethingelse}"`, the artifacts still end up being published using `VERSION_NAME` (i.e. settting `project.version` has no effect).

I'm pretty sure it's because of https://github.com/vanniktech/gradle-maven-publish-plugin/blob/3c9e994418e1fd078db56d7d5b7e58fa0e83cb02/src/main/kotlin/com/vanniktech/maven/publish/MavenPublishPom.kt#L33

gabrielittner:
closes #121

### DIFF
--- a/src/main/kotlin/com/vanniktech/maven/publish/MavenPublishConfigurer.kt
+++ b/src/main/kotlin/com/vanniktech/maven/publish/MavenPublishConfigurer.kt
@@ -47,7 +47,7 @@ internal class MavenPublishConfigurer(
 
   private fun configurePom(
     publication: MavenPublication,
-    groupId: String = project.group as String, // the plugin initially sets project.group to publishPom.groupId
+    groupId: String = project.group as String, // The plugin initially sets project.group to publishPom.groupId
     artifactId: String = publishPom.artifactId
   ) {
     publication.groupId = groupId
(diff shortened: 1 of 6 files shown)


src/main/kotlin/com/vanniktech/maven/publish/MavenPublishConfigurer.kt
```diff
@@ -47,12 +47,12 @@ internal class MavenPublishConfigurer(
 
   private fun configurePom(
     publication: MavenPublication,
-    groupId: String = publishPom.groupId,
+    groupId: String = project.group as String, // the plugin initially sets project.group to publishPom.groupId
```
vanniktech:
```suggestion
    groupId: String = project.group as String, // The plugin initially sets project.group to publishPom.groupId
```

src/main/kotlin/com/vanniktech/maven/publish/MavenPublishConfigurer.kt
```diff
@@ -47,12 +47,12 @@ internal class MavenPublishConfigurer(
...
     artifactId: String = publishPom.artifactId
   ) {
     publication.groupId = groupId
     publication.artifactId = artifactId
-    publication.version = publishPom.version
+    publication.version = project.version as String // the plugin initially sets project.version to publishPom.version
```
vanniktech:
```suggestion
    publication.version = project.version as String // The plugin initially sets project.version to publishPom.version
```

PR #127, opened 2020-03-29
update documentation and error message for closeAndReleaseRepository

gabrielittner:
We're defaulting `mavenPublish.nexus.groupId` to `GROUP` but it's often not. The `groupId` needs to be what you see in Sonatype, not the group of the library. Maybe we should rename the property to make it more clear.

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,6 +158,9 @@ mavenPublish {
     }
 }
 ```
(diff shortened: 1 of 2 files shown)


gabrielittner:
I think it's either the staging profile name or staging profile target on Sonatype, not sure which one of those two though
Screenshot 2020-03-29 at 21 03 03

PR #140, opened 2020-05-18
Use POM_INCEPTION_YEAR gradle property

### LINKED ISSUE #134: POM_INCEPTION_YEAR not implemented
The readme indicates that this plugin uses a config property `POM_INCEPTION_YEAR`, but on inspection it doesn't seem like anything in the code actually uses this property.  I noticed this when migrating a project from its old `gradle-mvn-push.gradle` file to this plugin, and the pom no longer has that field.

Not really a dealbreaker - I'm not sure who even cares about that pom field - but noteworthy nonetheless.

Would you be open to a PR implementing this?  Seems pretty simple to add, if desired.

benjamin-bader:
The project README indicates that a Gradle property `POM_INCEPTION_YEAR` can be set to specify a value for the similarly-named pom.xml `<inceptionYear/>` element. This property currently isn't used by the plugin.

This PR implements inception-year handling.

Fixes #134


vanniktech:
@gabrielittner feel free to merge if you also agree that everything is fine

PR #146, opened 2020-07-03
Remove android offline docs

### LINKED ISSUE #145: Android SDK no longer publishes docs
This line would need to be changed to just reference the android reference remotely

https://github.com/vanniktech/gradle-maven-publish-plugin/blob/master/src/main/kotlin/com/vanniktech/maven/publish/tasks/AndroidJavadocs.kt#L33

Otherwise android javadoc gen will fail on fresh SDK installs that no longer have this docs dir

ZacSweers:
Resolves #145


ZacSweers:
Got confirmation from the AGP folks that these are no longer published

ZacSweers:
Superseded by #147

PR #152, opened 2020-08-14
update to Gradle 6.6

gabrielittner:
The integration tests changed a bit because is published with metadata now. So the expected poms changes and we can generally check that a `.module` file gets created and published.

### DIFF
--- a/src/main/kotlin/com/vanniktech/maven/publish/nexus/CloseAndReleaseRepositoryTask.kt
+++ b/src/main/kotlin/com/vanniktech/maven/publish/nexus/CloseAndReleaseRepositoryTask.kt
@@ -1,10 +1,12 @@
 package com.vanniktech.maven.publish.nexus
 
 import org.gradle.api.DefaultTask
+import org.gradle.api.tasks.Input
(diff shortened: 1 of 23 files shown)


src/main/kotlin/com/vanniktech/maven/publish/nexus/CloseAndReleaseRepositoryTask.kt
```diff
@@ -1,10 +1,12 @@
 package com.vanniktech.maven.publish.nexus
 
 import org.gradle.api.DefaultTask
+import org.gradle.api.tasks.Input
 import org.gradle.api.tasks.TaskAction
 
 open class CloseAndReleaseRepositoryTask : DefaultTask() {
 
+  @Input
   lateinit var nexusOptions: NexusOptions
```
ZacSweers, line 10: Can these be inputs if they're not serializable?

gabrielittner: Missed that. I changed it to Nested and annotated NexusOptions

Issue #138, opened 2020-05-15
Add disableAutomaticPublicationCreation flag


gabrielittner:
Currently we have a lot of magic going on behind the scenes which makes it hard to deviate from the standard use cases. If the new `disableAutomaticPublicationCreation` flag is enabled we don't create any publication and don't create set up sources, javadoc/dokka tasks by default.

How manual configuration would look like:
`java`, `java-library`
```
publishing {
    publications {
        create<MavenPublication>("maven") {
            from(components["java"])
        }
    }
}

java {
  withSourcesJar()
  withJavadocJar()
}
```

`java-gradle-plugin`
```
java {
  withSourcesJar()
  withJavadocJar()
}
```

`com.android.library`
```
afterEvaluate {
    publishing {
        publications {
            create<MavenPublication>("maven") {
                from(components["release"])
            }
        }
    }
}

// new API by us until AGP offers something, exact name TBD
androidPublishCompat {
  withSourcesJar()
  withJavadocJar()
}
```

`org.jebtrains.kotlin.multiplatform`
```
// new API by us until Kotlin offers something, exact name TBD
kotlinMultiplatformPublishCompat {
  // sources jar already set up by default
  withJavadocJar()
}
```

As an example for something that wasn't supported directly before: `com.github.johnrengelman.shadow`
```
publishing {
    publications {
        create<MavenPublication>("maven") { publication ->
            project.shadow.component(publication)
        }
    }
}

java {
  withSourcesJar()
  withJavadocJar()
}
```

### What this plugin still does when the flag is on
- configure the pom for all publications (this would be a general behavior change, for java-library and Android, we're currently just touching the publication we create)
- configure signing
- some compatibility things like adding an empty sources jar to the `kotlinMultiplatform` publication until those things are fixed in the respective upstream repository

This generally brings us into the direction of just augmenting the default Gradle/AGP/Kotlin plugins with missing parts until they catch up, plus providing some nice defaults (maven central publishing, signing) and nicer configuration (pom through gradle properties).

vanniktech:
Sounds good!

gabrielittner:
Obsoleted by #173

PR #155, opened 2020-09-02
Dokka 1.4 compatibility

### LINKED ISSUE #150: Dokka 1.4.0-rc incompatibility
` Task with name 'dokka' not found in project ':mvrx'`
Unfortunately, the `gradle-maven-publish-plugin` is not compatible with the master or 1.4.0-rc branch of dokka. 
There is no single dokka task anymore. There are tasks like `dokkaHtml`, `dokkaJavadoc`, ..., etc.

Feel free to contact me in the Kotlin Slack if you need any help when trying to support new versions of dokka ☺️

gabrielittner:
fixes #150 

For now the logic will be:
- if there is just one `DokkaTask` take that
- otherwise use `dokkaHtml` (one of the default tasks created in 1.4)
- and the fallback is the current `dokka` task

I want to avoid adding an option until #138 is figured out.

This also changes the dependency ok dokka to `compileOnly`. Newer versions don't have an extra Android artifact anymore and we're less coupled to it. Users that only implicitly depended on it before will see an error and need to add it manually.

Branch to show that it works with 1.4 https://github.com/vanniktech/gradle-maven-publish-plugin/compare/dokka...dokka-1.4-rc

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -74,9 +74,8 @@ dependencies {
   api gradleApi()
   api "org.jetbrains.kotlin:kotlin-stdlib:$kotlinVersion"
 
-  implementation "org.jetbrains.dokka:dokka-gradle-plugin:$dokkaVersion"
-  implementation "org.jetbrains.dokka:dokka-android-gradle-plugin:$dokkaVersion"
-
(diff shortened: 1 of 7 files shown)


vanniktech (approved):
Really nice 👍

PR #165, opened 2020-09-12
set up Github Actions

gabrielittner:
@vanniktech Can you set up the secrets? For the pgp key we can also create a new one, in case the current one is used for other projects/things as well


.github/workflows/build.yml
```diff
@@ -0,0 +1,27 @@
...
+on: [push, pull_request]
+
+jobs:
+  build:
+    name: JDK ${{ matrix.java_version }}
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        java_version: [1.8]
```
ZacSweers, line 12: Do we care about adding others here eventually? Otherwise it can just be inlined to the action use

.github/workflows/build.yml
```diff
@@ -0,0 +1,27 @@
...
+      - name: Gradle Wrapper Validation
+        uses: gradle/wrapper-validation-action@v1
+
+      - name: Install JDK ${{ matrix.java_version }}
+        uses: actions/setup-java@v1
+        with:
+          java-version: ${{ matrix.java_version }}
+
+      - name: Build with Gradle
+        run: ./gradlew build --stacktrace
```
ZacSweers, line 27: Should we add `check` too?

.github/workflows/publish-release.yml
```diff
@@ -0,0 +1,33 @@
+name: Publish Release
```
ZacSweers, line 1: The workflow is 💯

.github/workflows/publish-snapshot.yml
```diff
@@ -0,0 +1,31 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Install JDK 8
+        uses: actions/setup-java@v1
+        with:
+          java-version: 1.8
+
+      - name: Retrieve version
```
ZacSweers, line 22: Nice

.github/workflows/publish-snapshot.yml
```diff
@@ -0,0 +1,31 @@
...
+        run: |
+          echo ::set-env name=VERSION_NAME::$(cat gradle.properties | grep -w "VERSION_NAME" | cut -d'=' -f2)
+
+      - name: Publish snapshot
+        run: ./gradlew uploadArchives --no-daemon --no-parallel
+        if: endsWith(env.VERSION_NAME, '-SNAPSHOT')
```
ZacSweers, line 28: Double nice

build.gradle
```diff
@@ -101,6 +101,12 @@ dependencies {
 
 sourceCompatibility = JavaVersion.VERSION_1_7
 
+signing {
+  if (hasProperty('SIGNING_PRIVATE_KEY')) {
+    useInMemoryPgpKeys(SIGNING_PRIVATE_KEY, "")
```
ZacSweers, line 106: Is this a Gradle API?

ZacSweers (approved):

.github/workflows/build.yml
gabrielittner: I wanted to add 11 and 14, but they are failing currently so I removed them for now.

build.gradle
gabrielittner: Yes, I found that through sqldelight 
https://docs.gradle.org/current/userguide/signing_plugin.html#sec:signatory_credentials

.github/workflows/build.yml
gabrielittner: `build` is `assemble` + `check` afaik

vanniktech:
>  For the pgp key we can also create a new one

Yep that's the case. I set this up years ago, do you happen to have instructions? Otherwise, I'll just google.

gabrielittner:
For creating the keypair and publishing the new public key you can follow these steps

To export the private key for the secret this command should work:
`gpg2 --export-secret-keys --armor KEY_ID`

vanniktech (approved):
Thanks. I've added the secrets. 🤞 sorry it took me longer

vanniktech (changes requested):
One thing, you need to delete the build step from the travis yml file that executes the deploy_snapshot

gabrielittner:
@vanniktech I deleted the travis file

vanniktech (approved):
🙈

PR #166, opened 2020-09-27
Add status badges

anggrayudi:
Adds status badges for:
* Latest stable version
* Build status

Screen Shot 2020-09-27 at 21 04 06


vanniktech (commented):
Thanks but no thanks. I'm not a fan of badges and all of my repos don't have any.

anggrayudi:
Ok then, I will decline this PR.

Issue #171, opened 2020-11-24
Eliminate property file usage


JakeWharton:
This seems to be a holdover from Chris' script, but it feels very out-of-place in modern Gradle.

Gradle has built-in mechanisms for specifying the group, artifact, and version. You're already required to use them in order to get peer dependencies working properly. These canonical sources should be used rather than a properties file.

Specifying all of the additional information for a POM feels like it should be done in an extension. This would allow project-local specification of name and description while using `subprojects { plugins.withId(..) { .. } }` for shared configuration such as URLs and developer.

ZacSweers:
we're planning to do an API revamp with breaking changes and raise the min API version, so this would be a good candidate for that 👍

Issue #154, opened 2020-09-01
Feature request: support Gradle's new Provider<PasswordCredentials> API


ZacSweers:
Talked with @gabrielittner about this, I plan to submit a PR. Creating an issue so I can link it in a TODO in our own code :) 

```kotlin
val credentials = providers.credentials(PasswordCredentials::class, "SonatypeNexusUsername")

mavenPublish {
  targets {
    getByName("uploadArchives") {
      credentials(credentials)
      // Or credentials("SonatypeNexusUsername")
    }
  }
}
```

vanniktech:
Sounds good, you two should take over this plugin.

gabrielittner:
With #213 NexusOptions is deprecated. We're using the credentials API to configure the publishing repository and are reusing the same properties for our `closeAndReleaseRepository` task

Issue #137, opened 2020-05-15
Deprecate targets API


gabrielittner:
We should deprecate the API to set create targets.

This is what we have:
```
mavenPublish {
  targets {
    myRepository {
      releaseRepositoryUrl = ""
      snapshotRepositoryUrl = ""
      repositoryUsername = ""
      repositoryPassword = ""
    }
  }
}
```

This is the standard Gradle API:
```
publishing {
    repositories {
        maven {
            url = ""
            credentials {
                username ""
                password ""
            }
        }
    }
}
```
The only advantage of our API is that we automatically switch between release and snapshot repository, but I think that's fine for additionally configured repositories. 

For our current defaults: We can and should still setup Sonatype OSS by default. Same behavior as right now with automatically switching between release and snapshot repositories. We can have a flag in our extension to disable adding it. The default maven local target can be removed, because Gradle automatically creates `publishToMavenLocal` and `publish<PubName>PublicationToMavenLocal` tasks.

As part of this I would also deprecate our `uploadArchives`, `installArchives`, `uploadArchivesRepoName` task aliases. There is no reason not to use `publish`, `publishAllPublicationsTo<RepositoryName>Repository` and `publishToMavenLocal` directly.

vanniktech:
You'd still be able to push to multiple / different repositories (bintray, mavenCentral) etc, right?

gabrielittner:
Yes, you just configure them through Gradle APIs instead of our own.

PR #173, opened 2020-11-28
[WIP] Modern Gradle Publish Plugin design document

gabrielittner:
Early draft of how the plugin should look like in the future. This includes issues #137, #138, #154 and #171.

For everything where we add an API specific to the project type (Android, Kotlin MPP etc.) I'll either open a bug on the respective project or will add a link to an existing one.

### DIFF
--- a/modern-publish-plugin.md
+++ b/modern-publish-plugin.md
@@ -1,0 +1,211 @@
+# Modern Gradle Publish Plugin
+
+
+## Goals
+
+- less magic configuration, more explicit APIs
+- provide APIs for everything represented through properties
+- remove or deprecate APIs that are already provided by Gradle
+- continue providing a way to get started with minimal configuration
+
+## New Structure
+
+For compatibility purposes the current `com.vanniktech.maven.publish` plugin
+behavior is kept as it is, with the exception of APIs that we decide to remove. 
+Most of the goals are solved through a set of new plugins for specific 
+platforms. These plugins provide APIs that replace the magic setup mechanisms
+of the current plugin. To avoid having to maintain two implentations the old
+plugin will internally just apply the new plugins and use their public APIs. 
+Over time we can decide if `com.vanniktech.maven.publish` should be deprecated
+or removed.
+
+### Plugins
+
+Specific plugins for each supported project type:
+- `com.vanniktech.maven.publish.java`
+- `com.vanniktech.maven.publish.gradle`
+- `com.vanniktech.maven.publish.android`
+- `com.vanniktech.maven.publish.kotlin.js`
+- `com.vanniktech.maven.publish.kotlin.mpp`
+
+There are two reasons for going with individual plugins. 
+
+
+On top of that there will be a `com.vanniktech.maven.publish.base` plugin that 
+gets applied by all of these and the existing `com.vanniktech.maven.publish` as 
+described above.
+
+#### `com.vanniktech.maven.publish.base`
+
+- applies `maven-publish` Gradle plugin
+- provides the API that is shared for all project types
+- can be used for shared configurations using `plugins.withId(...)` in projects with mixed types
+
+```
+mavenPublish {
+  # applies signing plugin and tells Gradle to sign all publications
+  signAllPublications()
+  # adds the Sonatype release/snapshot repo as targets based on version
+  # uses Gradle credentials API for username/password
+  # automatically calls signAllPublications()
+  # sets nexus.baseUrl
+  publishToMavenCentral()
+
+  # might not be needed as it defaults to the mutable `project.getGroup()`
+  groupId = "com.example"
+  # might not be needed as it defaults to the mutable `project.getGroup()`
+  version = "1.0.0"
+  # Base because their might be multiple artifacts produced by one module.
+  # Without this the arifact id would always be based on `project.getName()`.
+  # Implementation in Kotlin MPP is kind of hacky using replace so it might be
+  # best to just not support it. How do we handle compatibility?
+  baseArtifactId = "library"
+
+  # Same API as Gradle's, can probably just use their interfaces/classes
+  # Will be applied to all publications
+  pom {
+    name = "My Library"
+    description = "A description of what my library does."
+    inceptionYear = "2020"
+    url = "https://github.com/username/mylibrary/"
+    
+    scm {
+        connection = "scm:git:git://github.com/username/mylibrary.git"
+        developerConnection = "scm:git:ssh://git@github.com/username/mylibrary.git"
+        url = "https://github.com/username/mylibrary/"
+    } 
+    
+    licenses {
+        license {
+            name = "The Apache Software License, Version 2.0"
+            url = "https://www.apache.org/licenses/LICENSE-2.0.txt"
+            distribution = "repo"
+        }
+    }
+    developers {
+        developer {
+            id = "username"
+            name = "User Name"
+            url = "https://github.com/username/"
+        }
+    }
+  }
+
+  # Configure closeAndReleaseRepository
+  nexus {
+     baseUrl = "https://your_nexus_instance"
+     stagingProfile = "net.example"
+     # Start supporting the Gradle credentials API
+     respositoryUserName = "username"
+     respositoryPassword = "password"
+  }
+}
+```
+
+#### `com.vanniktech.maven.publish.java`
+
+Meant to be used with `java`, `java-library`, `org.jetbrains.kotlin.jvm`. 
+
+Pure Gradle configuration:
+```
+publishing {
+    publications {
+        create<MavenPublication>("maven") {
+            from(components["java"])
+        }
+    }
+}
+
+java {
+  withSourcesJar()
+  withJavadocJar()
+}
+```
+
+The plugin doesn't need to offer anything for sources and javadocs jars. It will
+simply create a publication using `components["java"]`.
+
+#### `com.vanniktech.maven.publish.gradle`
+
+Meant to be used with `java-gradle-plugin`.
+
+Pure Gradle configuration:
+```
+java {
+  withSourcesJar()
+  withJavadocJar()
+}
+```
+
+Gradle automatically adds a publication when both `java-gradle-plugin` and 
+`maven-publish` are applied. There is no need for our own APIs, so this plugin
+does not have any functionality of it's own.
+
+#### `com.vanniktech.maven.publish.android`
+
+Meant to be used with `com.android.library`.
+
+Pure Gradle configuration:
+```
+afterEvaluate {
+    publishing {
+        publications {
+            create<MavenPublication>("maven") {
+                from(components["release"])
+            }
+        }
+    }
+}
+```
+
+
+There is no API available to create source and javadoc jars for Android 
+projects. The plugin will provide an API to configure those as well as 
+which variant to publish. These properties/methods can be removed/deprecated once
+AGP offers it's own methods and adds multi variant publishing.
+
+```
+androidPublishCompat {
+  androidVariantToPublish = "release"
+  withSourcesJar()
+  withJavadocJar()
+}
+```
+
+#### `com.vanniktech.maven.publish.kotlin.js`
+
+Meant to be used with `org.jetbrais.kotlin.js`.
+
+Pure Gradle configuration:
+```
+TODO
+```
+
+TODO
+
+#### `com.vanniktech.maven.publish.kotlin.mpp`
+
+Meant to be used with `org.jetbrais.kotlin.multiplatform`.
+
+Pure Gradle configuration: n/a
+
+Like `java-gradle-plugin` the Kotlin mpp plugin sets up a publication 
+automatically. It also adds source jars automatically. This plugin adds
+an API to add javadoc publishing. There still is a `withSourcesJar()` which will
+add empty source jars to all publications missing one, this can be removed once
+the behavior is fixed on the Kotlin side.
+
+```
+kotlinMultiplatformPublishCompat {
+  withSourcesJar()
+  withJavadocJar()
+}
+```
+
+#### `com.vanniktech.maven.publish`
+
+- calls `publishToMavenCentral()`
+- sets `groupId`, `version` and `baseArtifactId` based on Gradle properties
+- configures `pom` based on Gradle properties
+- applies project type plugin automatically based on other applied plugins
+- calls all API methods of the project type plugin
--- a/modern-publish-plugin.md
+++ b/modern-publish-plugin.md
@@ -46,6 +46,7 @@ described above.
 ```
 mavenPublish {
   # applies signing plugin and tells Gradle to sign all publications
+  # if the version ends with -SNASHOT signing will be skipped automatically if credentials are missing
   signAllPublications()
   # adds the Sonatype release/snapshot repo as targets based on version
   # uses Gradle credentials API for username/password
--- a/modern-publish-plugin.md
+++ b/modern-publish-plugin.md
@@ -25,11 +25,13 @@ Specific plugins for each supported project type:
 - `com.vanniktech.maven.publish.java`
 - `com.vanniktech.maven.publish.gradle`
 - `com.vanniktech.maven.publish.android`
+- `com.vanniktech.maven.publish.kotlin.jvm`
 - `com.vanniktech.maven.publish.kotlin.js`
 - `com.vanniktech.maven.publish.kotlin.mpp`
 
-There are two reasons for going with individual plugins. 
-
+The main motivation for separating these, is to provide specific APIs for each project
+without running into race conditions when something like `plugins.withId` is used or 
+requiring ordering. Another reason is that it allows us to nicely separate the code. 
 
 On top of that there will be a `com.vanniktech.maven.publish.base` plugin that 
 gets applied by all of these and the existing `com.vanniktech.maven.publish` as 
@@ -48,18 +50,8 @@ mavenPublish {
   # adds the Sonatype release/snapshot repo as targets based on version
   # uses Gradle credentials API for username/password
   # automatically calls signAllPublications()
-  # sets nexus.baseUrl
+  # sets nexus.baseUrl, nexus.respositoryUserName, nexus.respositoryPassword
   publishToMavenCentral()
-
-  # might not be needed as it defaults to the mutable `project.getGroup()`
-  groupId = "com.example"
-  # might not be needed as it defaults to the mutable `project.getGroup()`
-  version = "1.0.0"
-  # Base because their might be multiple artifacts produced by one module.
-  # Without this the arifact id would always be based on `project.getName()`.
-  # Implementation in Kotlin MPP is kind of hacky using replace so it might be
-  # best to just not support it. How do we handle compatibility?
-  baseArtifactId = "library"
 
   # Same API as Gradle's, can probably just use their interfaces/classes
   # Will be applied to all publications
@@ -102,7 +94,7 @@ mavenPublish {
 }
 ```
 
-#### `com.vanniktech.maven.publish.java`
+#### `com.vanniktech.maven.publish.java` & `com.vanniktech.maven.publish.kotlin.jvm`
 
 Meant to be used with `java`, `java-library`, `org.jetbrains.kotlin.jvm`. 
 
@@ -177,11 +169,30 @@ androidPublishCompat {
 Meant to be used with `org.jetbrais.kotlin.js`.
 
 Pure Gradle configuration:
-```
-TODO
+```publishing {
+    publications {
+        create<MavenPublication>("maven") {
+            from(components["kotlin"])
+            artifact(project.tasks.named("kotlinSourcesJar"))
+        }
+    }
+}
 ```
 
-TODO
+
+There is no API available to create source and javadoc jars for Kotlin JS
+projects, but there is at least a task for sources. The plugin will provide 
+an API to configure those. These properties/methods can be removed/deprecated once
+Kotlin offers it's own methods and adds multi variant publishing.
+
+```
+androidPublishCompat {
+  androidVariantToPublish = "release"
+  withSourcesJar()
+  withJavadocJar()
+}
+```
+
 
 #### `com.vanniktech.maven.publish.kotlin.mpp`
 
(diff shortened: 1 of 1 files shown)


modern-publish-plugin.md
```diff
@@ -0,0 +1,211 @@
...
+or removed.
+
+### Plugins
+
+Specific plugins for each supported project type:
+- `com.vanniktech.maven.publish.java`
+- `com.vanniktech.maven.publish.gradle`
+- `com.vanniktech.maven.publish.android`
+- `com.vanniktech.maven.publish.kotlin.js`
+- `com.vanniktech.maven.publish.kotlin.mpp`
```
gabrielittner, line 29: These are getting pretty long, maybe there are shorter names that are still good?

modern-publish-plugin.md
```diff
@@ -0,0 +1,211 @@
...
+     # Start supporting the Gradle credentials API
+     respositoryUserName = "username"
+     respositoryPassword = "password"
+  }
+}
+```
+
+#### `com.vanniktech.maven.publish.java`
+
+Meant to be used with `java`, `java-library`, `org.jetbrains.kotlin.jvm`.
```
gabrielittner, line 107: Should there be a `com.vanniktech.maven.publish.kotlin.jvm`  that's just an alias to `com.vanniktech.maven.publish.java`? For consistency with the other Kotlin plugins.

modern-publish-plugin.md
```diff
@@ -0,0 +1,211 @@
...
+described above.
+
+#### `com.vanniktech.maven.publish.base`
+
+- applies `maven-publish` Gradle plugin
+- provides the API that is shared for all project types
+- can be used for shared configurations using `plugins.withId(...)` in projects with mixed types
+
+```
+mavenPublish {
```
gabrielittner, line 45: Should we keep `mavenPublish` as extension name?

modern-publish-plugin.md
```diff
@@ -0,0 +1,211 @@
...
+
+  # might not be needed as it defaults to the mutable `project.getGroup()`
+  groupId = "com.example"
+  # might not be needed as it defaults to the mutable `project.getGroup()`
+  version = "1.0.0"
+  # Base because their might be multiple artifacts produced by one module.
+  # Without this the arifact id would always be based on `project.getName()`.
+  # Implementation in Kotlin MPP is kind of hacky using replace so it might be
+  # best to just not support it. How do we handle compatibility?
+  baseArtifactId = "library"
```
gabrielittner, line 62: Should we have `groupId`, `version` and `baseArtifactId` extension properties? 
Gradle itself has an API to set/change them on a publication, so it's not always limited to the project name, but it gets hacky with publications not created by us.

modern-publish-plugin.md
```diff
@@ -0,0 +1,211 @@
...
+
+```
+mavenPublish {
+  # applies signing plugin and tells Gradle to sign all publications
+  signAllPublications()
+  # adds the Sonatype release/snapshot repo as targets based on version
+  # uses Gradle credentials API for username/password
+  # automatically calls signAllPublications()
+  # sets nexus.baseUrl
+  publishToMavenCentral()
```
gabrielittner, line 52: Should this be used as an indicator by project plugins to automatically enable source/javadoc artifacts or is it just the magic `com.vanniktech.maven.publish` plugin that adds those automatically?

modern-publish-plugin.md
```diff
@@ -0,0 +1,211 @@
...
+- can be used for shared configurations using `plugins.withId(...)` in projects with mixed types
+
+```
+mavenPublish {
+  # applies signing plugin and tells Gradle to sign all publications
+  signAllPublications()
+  # adds the Sonatype release/snapshot repo as targets based on version
+  # uses Gradle credentials API for username/password
+  # automatically calls signAllPublications()
+  # sets nexus.baseUrl
```
gabrielittner, line 51: Also username and password

modern-publish-plugin.md
```diff
@@ -0,0 +1,211 @@
...
+```
+
+
+There is no API available to create source and javadoc jars for Android 
+projects. The plugin will provide an API to configure those as well as 
+which variant to publish. These properties/methods can be removed/deprecated once
+AGP offers it's own methods and adds multi variant publishing.
+
+```
+androidPublishCompat {
```
gabrielittner, line 168: Better name?

modern-publish-plugin.md
```diff
@@ -0,0 +1,211 @@
...
+Pure Gradle configuration: n/a
+
+Like `java-gradle-plugin` the Kotlin mpp plugin sets up a publication 
+automatically. It also adds source jars automatically. This plugin adds
+an API to add javadoc publishing. There still is a `withSourcesJar()` which will
+add empty source jars to all publications missing one, this can be removed once
+the behavior is fixed on the Kotlin side.
+
+```
+kotlinMultiplatformPublishCompat {
```
gabrielittner, line 199: Better name?

modern-publish-plugin.md
```diff
@@ -0,0 +1,211 @@
...
+  withSourcesJar()
+  withJavadocJar()
+}
+```
+
+#### `com.vanniktech.maven.publish`
+
+- calls `publishToMavenCentral()`
+- sets `groupId`, `version` and `baseArtifactId` based on Gradle properties
+- configures `pom` based on Gradle properties
```
gabrielittner, line 209: Should we add a deprecation message when reading `gradle.properties` or do we want to support them long term here?

modern-publish-plugin.md
gabrielittner: Maybe it could be a parameter on this method. That way we get easy configurability without forcing it.

modern-publish-plugin.md
```diff
@@ -0,0 +1,211 @@
...
+### Plugins
+
+Specific plugins for each supported project type:
+- `com.vanniktech.maven.publish.java`
+- `com.vanniktech.maven.publish.gradle`
+- `com.vanniktech.maven.publish.android`
+- `com.vanniktech.maven.publish.kotlin.js`
+- `com.vanniktech.maven.publish.kotlin.mpp`
+
+There are two reasons for going with individual plugins.
```
ZacSweers, line 31: What are the two reasons?

modern-publish-plugin.md
vanniktech: I'm fine with these names.

modern-publish-plugin.md
vanniktech: I would.

modern-publish-plugin.md
vanniktech: How about we start without them and if required, we add them?

modern-publish-plugin.md
vanniktech: Does not 99% of the people want to push to maven central?

modern-publish-plugin.md
vanniktech: Yup

modern-publish-plugin.md
vanniktech: Can't we use androidPublish?

modern-publish-plugin.md
```diff
@@ -0,0 +1,211 @@
...
+
+TODO
+
+#### `com.vanniktech.maven.publish.kotlin.mpp`
+
+Meant to be used with `org.jetbrais.kotlin.multiplatform`.
+
+Pure Gradle configuration: n/a
+
+Like `java-gradle-plugin` the Kotlin mpp plugin sets up a publication
```
vanniktech, line 192:
```suggestion
Like `java-gradle-plugin` the Kotlin MPP plugin sets up a publication 
```

modern-publish-plugin.md
vanniktech: We could output one warning for all of them. That way we encourage upgrading without spamming the entire console for every single property.

modern-publish-plugin.md
gabrielittner: Yes, I still think it should be opt-in (so you need to something to enable it), but then it should be as effortless as possible. The reason for making it opt-in is to avoid having magic in the base plugins. I think having 2 parameters here to disable source/javadoc jar might be the best compromise. Just calling `publishToMavenCentral()` saves you from most config but it's still possible to enable it without automatically enabling javadoc/source jars (theoretically it's possible to have that for central).

modern-publish-plugin.md
gabrielittner: Sure, the main reason why I added was compat was because it's meant to be temporary until AGP handles this on it's own.

modern-publish-plugin.md
gabrielittner: - In general it allows us to separate the code better
- The main reason is that even with using `plugins.withId` and `afterEvaluate` there will be weird cases when plugins are defined in a certain order. For example let's say you define the publish plugin and then the android plugin and you have a `plugins.withId("...publish")` to call the `androidPublishCompat` API. The publish plugin would add the API only after it detects the Android plugin but that happens after the user's `plugins.withId`.

asarkar:
One thing that I don't understand is what does this plugin do that [Maven Publish Plugin](https://docs.gradle.org/current/userguide/publishing_maven.html) doesn't? Also, given that [JCenter and Bintray are going to be decommissioned](https://jfrog.com/blog/into-the-sunset-bintray-jcenter-gocenter-and-chartcenter/), people are going to be looking for alternatives, and any API changes preferably should be flushed out quickly, without the quintessential months or years of delay that usually plagues OSS.

vanniktech (approved):

modern-publish-plugin.md
```diff
@@ -0,0 +1,222 @@
...
+#### `com.vanniktech.maven.publish.base`
+
+- applies `maven-publish` Gradle plugin
+- provides the API that is shared for all project types
+- can be used for shared configurations using `plugins.withId(...)` in projects with mixed types
+
+```
+mavenPublish {
+  # applies signing plugin and tells Gradle to sign all publications
+  signAllPublications()
```
martinbonnin, line 49: Can we have more options there?

```
// Uses the gradle properties as explained in https://docs.gradle.org/current/userguide/signing_plugin.html#sec:signatory_credentials
signAllPublications()
// Uses in-memory keys which is more convenient to hold in an env variable in CI
signAllPublicationsWithInMemoryKeys(key, password)

// Allow to easily disable signing if developers do not have the secrets or in order to save some time
signingIsEnabled = ... //System.genv(SINGING_KEY) != null


```

modern-publish-plugin.md
```diff
@@ -0,0 +1,222 @@
...
+        developer {
+            id = "username"
+            name = "User Name"
+            url = "https://github.com/username/"
+        }
+    }
+  }
+
+  # Configure closeAndReleaseRepository
+  nexus {
```
martinbonnin, line 87: What about merging this with `publishToMavenCentral` ? Both of them require the sonatype username/password and this would allow to do fancy stuff like ensuring everything gets uploaded to the same staging repo:

```
addNexusRepository(
  baseUrl = ..
  username = ..
  password = ..
  stagingProfile = .. // if null will use the first stagingProfile
  repositoryId = .. // if null will create a new staging repository
)

addMavenCentralRepository() // a shortcut for addNexusRepository with baseUrl already configured
```

modern-publish-plugin.md
```diff
@@ -0,0 +1,222 @@
...
+
+### Plugins
+
+Specific plugins for each supported project type:
+- `com.vanniktech.maven.publish.java`
+- `com.vanniktech.maven.publish.gradle`
+- `com.vanniktech.maven.publish.android`
+- `com.vanniktech.maven.publish.kotlin.jvm`
+- `com.vanniktech.maven.publish.kotlin.js`
+- `com.vanniktech.maven.publish.kotlin.mpp`
```
martinbonnin, line 30: Another way to look at this would be to have one single plugin with `compileOnly` dependencies to all the aforementioned ones and let the user call appropriate methods later on:

```
mavenPublish {
   ...
   // creates a publication based on the "release" variant and empty javadoc/sources jars
   withAndroidPublications {
      // here there is a MavenPublication receiver to override behaviour if needed
      // e.g. if there is a sourcesJar available, one can do
      artifact(sourcesJar)
   }
} 
```

or in a kotlinMpp project:

```
mavenPublish {
   ...
   // does not create any publication as they are created by the kotlin plugin
   // Adds empty javadoc
   withKotlinMppPublications()
} 
```

modern-publish-plugin.md
JavierMeep: Avoid signing is important because snapshots doesnt require it and time is gold.

modern-publish-plugin.md
JavierMeep: For a KMP library, only mpp plugin is necessary or if it has an android target, android and mpp are needed?

martinbonnin: @JavierMeep you'll need both `com.android.library` and `org.jetbrains.kotlin.multiplatform` in a MPP project using Android.

`org.jetbrains.kotlin.multiplatform` should create all publications though. IIRC, it misses javadoc (for all publications) and sources (for the metatdata one, maybe for android too)

JavierMeep:
Use case: Multiplatform library which has Windows and macOS artifacts.

With `maven-publish` + GitHub Actions we got two repositories, one for each matrix (macos and windows). If there is a way to indicate the repositoryID like @martinbonnin mentioned before, we should be able to generate only one repo and close it:

1. Create the repo and, in some way, share the ID of this repo with the next steps/jobs
2. Publish all artifacts to that repo from macos and windows, this requires the previous job/step to be finished (macos will be uploding the artifacts at same time as Windows, there is no problem with that).
3. Close/Release the repo with the ID mentioned before, which requires the previous job/step to be finished

JavierSegoviaCordoba:
Use case: Multiplatform library which has Windows and macOS artifacts.

With maven-publish + GitHub Actions we got two repositories, one for each matrix (macos and windows). If there is a way to indicate the repositoryID like @martinbonnin mentioned before, we should be able to generate only one repo and close it:

1. Create the repo and, in some way, share the ID of this repo with the next steps/jobs
2. Publish all artifacts to that repo from macos and windows, this requires the previous job/step to be finished (macos will be uploding the artifacts at same time as Windows, there is no problem with that).
3. Close/Release the repo with the ID mentioned before, which requires the previous job/step to be finished

modern-publish-plugin.md
JavierSegoviaCordoba: I know, but I am talking about these:

- `com.vanniktech.maven.publish.android`
- `com.vanniktech.maven.publish.kotlin.mpp`

I think a kmp library with Android target should use only one plugin vanniktech plugin.

modern-publish-plugin.md
gabrielittner: I would keep the current behavior of not trying to sign when the current versions ends with `-SNAPSHOT`. If you don't want to sign non snapshot artifacts then you could also just do `if (conditionForSigning) signAllPublications()`, I don't think we need to offer an extra option here.

As for `signAllPublicationsWithInMemoryKeys`, I'm a bit split on this. It would be a bit shorter
```
mavenPublish {
  signAllPublications()
}

signing {
  if (project.hasProperty('SIGNING_PRIVATE_KEY') && project.hasProperty('SIGNING_PASSWORD')) {
    useInMemoryPgpKeys(project.getProperty('SIGNING_PRIVATE_KEY'), project.getProperty('SIGNING_PASSWORD'))
  }
}
```
vs 
```
mavenPublish {
  if (project.hasProperty('SIGNING_PRIVATE_KEY') && project.hasProperty('SIGNING_PASSWORD')) {
    signAllPublicationsWithInMemoryKeys(project.getProperty('SIGNING_PRIVATE_KEY'), project.getProperty('SIGNING_PASSWORD'))
  }
}
```
but also not too much. I could see value if we just have expected property names and the API would be `signAllPublicationsWithInMemoryKeys()` (again this would only make signing required for non SNAPSHOTS). While we are going away from Gradle properties they are still the best solution for handling credentials which this falls under. I have also opened an issue with Gradle to support the credentials API for signing https://github.com/gradle/gradle/issues/16056

modern-publish-plugin.md
gabrielittner: At the very least `publishToMavenCentral` will configure nexus: https://github.com/vanniktech/gradle-maven-publish-plugin/pull/173/files#diff-6a981871304a5f63c9fda36a805d124700fa5632c907f1c8e50bd91303863c16R53 So you wouldn't need to call it yourself unless you want to set a `stagingProfile`, but maybe that can be an optional parameter.

When I started writing the code for the base extension, I was unsure about how to handle `nexus`. In general I don't want to duplicate a perfectly fine Gradle API again (adding repositories). I also don't know how much of a use case `closeAndReleaseRepository` has outside of Sonatype OSSRH. Does someone who self hosts a Nexus instance use staging repositories for publishing or would they just publish directly to the repository. Maybe we leave out the API for now on base and see if someone needs it.

modern-publish-plugin.md
```diff
@@ -0,0 +1,222 @@
...
+kotlinMultiplatformPublishCompat {
+  withSourcesJar()
+  withJavadocJar()
+}
+```
+
+#### `com.vanniktech.maven.publish`
+
+- calls `publishToMavenCentral()`
+- sets `groupId`, `version` and `baseArtifactId` based on Gradle properties
```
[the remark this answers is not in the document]
gabrielittner: I agree which is why we don't have anything for these in the new API see the comments here https://github.com/vanniktech/gradle-maven-publish-plugin/pull/173#discussion_r532041099. This line is just for the old (current) plugin to keep it's current behavior like it does for other properties. Eventually this can get deprecated as well. We already supported setting group and version on the project yourself for a while and with 0.14.0 we removed the requirement to set these 3 gradle properties.

modern-publish-plugin.md
martinbonnin: > If you don't want to sign non snapshot artifacts then you could also just do `if (conditionForSigning) signAllPublications()`, I don't think we need to offer an extra option here.

Agreed that putting the `signAllPublications` under a `if` condition is simpler, no need for `signingIsEnabled()` 👍 

> While we are going away from Gradle properties they are still the best solution for handling credentials which this falls under.

I'd have to respectfully disagree here. I find environment variables to be easier in many circumstances. Mainly because they are super easy to setup in CI as well as on the dev machine.

modern-publish-plugin.md
martinbonnin: >  I also don't know how much of a use case closeAndReleaseRepository has outside of Sonatype OSSRH. 

I guess OSSRH is the main use case.

> So you wouldn't need to call it yourself unless you want to set a stagingProfile, but maybe that can be an optional parameter.

Yup, something like `publishToMavenCentral(stagingProfile: String? = null)`

I think there is a lot of value in the pluging avoiding split staging repositories though, i.e having `publishToMavenCentral` declare a repository with `url=`https://oss.sonatype.org/service/local/staging/deployByRepositoryId/$repositoryId`. This has been a pain point for many users.

All in all, something like this:

```
// adds a new ossrh repository that creates a new staging repository for this build
// @param stagingProfile: the staging profile to use. If null, will use the first one or fail if there is more than one
// registers a closeAndReleaseRepository task that will close and release the newly created staging repository (or fail if more than 1 repository are open)
publishToMavenCentral(stagingProfile: String? = null)
```

We can refine later by having the repositoryId an output property so that it can be reused. But avoiding the split repositories would be a nice first step.

modern-publish-plugin.md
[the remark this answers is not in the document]
JavierSegoviaCordoba: Sorry, I was reading the PR from mobile, I missed some things 😢

modern-publish-plugin.md
gabrielittner: You can create gradle properties by prefixing an environment variable with `ORG_GRADLE_PROJECT_`. For example you can have `ORG_GRADLE_PROJECT_mavenCentralRepositoryUsername` as an env var for the `mavenCentralRepositoryUsername` property

gabrielittner: We're doing that here for example https://github.com/vanniktech/gradle-maven-publish-plugin/blob/master/.github/workflows/publish-release.yml#L25-L28

martinbonnin: TBH, this is exactly the part that gets me confused every time 😅 . Gradle properties come with a lot of cognitive load (project vs system, ~/.gradle.properties or project/gradle.properties or env variables, does it have to be upper or lower case, etc, etc.....). I've read https://docs.gradle.org/current/userguide/build_environment.html multiple times and I still can't wrap my head around all the possibilities opened by gradle.properties. That why I usually prefer to avoid them. 

All in all, I think I'd like this plugin to be opiniated and hide the overwhelming options providing by Gradle under a simple API. I think having an api that takes two string parameters (armoured key + password) would be a nice default. That wouldn't prevent someone who wants to sign on a Yubikey to go read the Gradle Signing documentation and configure signing manually using gpg but I think that'd make it easier for beginners.

gabrielittner: Setting the sonatype username and password already happens through Gradle properties though (using Gradle's credential API). Would adding 2 more properties for in memory signing really make a difference? The default not in memory signing also uses 3 Gradle properties. I'll open a PR today we can discuss more there

modern-publish-plugin.md
gabrielittner: Would we even need `stagingProfile` if we create the staging repository through the API? The API seems to require a profile id which would be something different. So maybe it makes sense to leave this out for now.

martinbonnin: Sorry I meant `stagingProfileId`, not `stagingProfile`, these are the same thing that you can see at https://oss.sonatype.org/#stagingProfiles;$stagingProfileId

The more I think about this, the more I think 2 versions would be nice actually 😅 :

```
// Creates a new staging repo using the given stagingProfileId or the first one found if null
publishToMavenCentral(stagingProfileId: String? = null)
// Does not create a new staging repo but reuses one that is already created
publishToMavenCentralWithStagingRepository(stagingRepositoryId: String)
```

I'd like to make a small Github action that creates a staging repository. If that works, using the 2nd version up there would allow to have all artifacts uploaded to the same staging repository, for an exemple in MPP  builds spread on multiple VMs.

gabrielittner:
Closing this. All major parts have been implemented and will be available as a base plugin in 0.15.0. The platform approach implemented in #212 differs from what was described here, exchanging a huge set of plugins for a simple API. Any follow up can be done through regular issues.

PR #176, opened 2021-01-21
Adding optional subkey id

andretietz:
common use case, to use a "sign only" subkey.

Using it like this atm:
https://github.com/andretietz/retroauth/blob/master/gradle/publish.gradle#L16


gabrielittner:
Thanks for the contribution. We have a new way to set up in memory signing in `0.15.0`, the README update for that is in #227

PR #178, opened 2021-02-06
some cleanups

gabrielittner:
- Removed `Configurer` interface, we used to have 2 implementations but now there is only. I also think that the implementation won't survive the upcoming refactorings.
- In preparation of the new DSL it's now not required to set group id, artifact id and version properties anymore. The fallback are the respective project properties that Gradle uses as well.
- ~Bumped Gradle, Kotlin and Moshi~
- Deduplicated the integration test setup, by sharing settings.gradle files and the repository declaration.


gabrielittner:
I will extract the version update because there are some behavior changes

src/main/kotlin/com/vanniktech/maven/publish/MavenPublishPom.kt
```diff
@@ -29,9 +29,9 @@ internal data class MavenPublishPom(
   internal companion object {
     @JvmStatic
     fun fromProject(project: Project) = MavenPublishPom(
-        project.findMandatoryProperty("GROUP"),
-        project.findMandatoryProperty("POM_ARTIFACT_ID"),
-        project.findMandatoryProperty("VERSION_NAME"),
+        project.findOptionalProperty("GROUP") ?: project.group.toString(),
```
vanniktech, line 32: Is there internal validation within the constructor that fails on empty gruops / names etc

vanniktech (approved):

src/main/kotlin/com/vanniktech/maven/publish/MavenPublishPom.kt
gabrielittner: Gradle has a default value for the group if it's not set explicitly: https://docs.gradle.org/current/javadoc/org/gradle/api/Project.html#getGroup--

vanniktech (approved):

PR #179, opened 2021-02-06
update Kotlin to 1.4.30

gabrielittner:
Kotlin fixed the missing sources jar on their side, which is breaking us. I've changed our fix to only run if there is no sources artifact yet, so we are keeping compatibility with older Kotlin versions.

https://youtrack.jetbrains.com/issue/KT-44298


vanniktech (approved):
Nice!

PR #180, opened 2021-02-06
bump Gradle to 6.8.2

gabrielittner:
Just opening this to show https://github.com/gradle/gradle/issues/15731 is still happening


gabrielittner:
Changed how we set the `artifactId` to fix the Gradle issue caused by `afterEvaluate`. However I needed to add a workaround for mpp projects because for them it doesn't work without `afterEvaluate`

src/integrationTest/kotlin/com/vanniktech/maven/publish/MavenPublishPluginIntegrationTest.kt
```diff
@@ -162,6 +162,8 @@ class MavenPublishPluginIntegrationTest(
     assertThat(result.task(":$uploadArchivesTargetTaskName")?.outcome).isEqualTo(SUCCESS)
     assertThat(result.task(":dokka")?.outcome).isEqualTo(SUCCESS)
 
+    repoFolder.walkTopDown().forEach { println(it) }
```
vanniktech, line 165: Do we still need / want this?

gabrielittner: No, missed that one 7623d50

vanniktech (approved):

PR #182, opened 2021-02-07
simplify Nexus task setup and option to specify repository

gabrielittner:
This should improve our handling of staging repositories. By default it will behave like before, if you specify a `stagingProfile` we'll try to find a repository matching it. However `stagingProfile` is not required anymore. When it's not set then we'll just look at all staging repositories, if there is just one we'll close and release it. In the most common scenario this should be enough. An additional improvement is that if more than one is found and you know which one you want to release (or all of them) you can call `./gradlew closeAndReleaseRepository --repository comexample-123` to do so. The exceptions for not finding a repository will also print all of them.

Additionally makes all the nexus classes `internal`.


src/main/kotlin/com/vanniktech/maven/publish/nexus/CloseAndReleaseRepositoryTask.kt
```diff
@@ -3,34 +3,28 @@ package com.vanniktech.maven.publish.nexus
...
 import org.gradle.api.tasks.Nested
 import org.gradle.api.tasks.TaskAction
+import org.gradle.api.tasks.options.Option
 
 open class CloseAndReleaseRepositoryTask : DefaultTask() {
 
   @Nested
   lateinit var nexusOptions: NexusOptions
 
+  @Option(option = "repository", description = "Specify which staging repository to close and release.")
```
vanniktech, line 13: TIL

src/main/kotlin/com/vanniktech/maven/publish/nexus/Nexus.kt
```diff
@@ -8,7 +8,13 @@ import retrofit2.Retrofit
 import retrofit2.converter.moshi.MoshiConverterFactory
 import java.io.IOException
 
-class Nexus(username: String, password: String, val stagingProfile: String, baseUrl: String) {
+internal class Nexus(
+  baseUrl: String,
+  username: String,
+  password: String,
+  private val stagingProfile: String?,
+  private val stagingRepository: String?
```
vanniktech, line 16:
```suggestion
  private val stagingRepository: String?,
```

vanniktech (approved):

martinbonnin:
Nice 👍 . Something else that would be handy is explicitly creating a staging repository and uploading all files to it like https://github.com/marcphilipp/nexus-publish-plugin does. 

For multiplatform scenarios, that'll allow having different CI vms upload artifacts in parallel. 

One step further, maybe allow to "remember" and share that repositoryId to all the vms in the current build. This way, mac, windows, linux vms will all upload to the same repo, making the release more atomic.

PR #183, opened 2021-02-07
deprecate old target API

### LINKED ISSUE #137: Deprecate targets API

gabrielittner:
- deprecates the target API
- add deprecation message to custom `uploadArchives`/`installArchives` tasks
- deprecated repository configuration through env vars and properties
- deprecated old username/password properties in favor of `mavenCentralRepositoryUsername` and `mavenCentralRepositoryPassword` which is what these properties would look like using the credential API (#154) assuming that in the future we create a default repository called `mavenCentral` using Gradle's API

Closes #137


vanniktech (commented):
Can we already add an example to the README how you'd use the Gradle API to configure the targets?

vanniktech (approved):

PR #186, opened 2021-02-10
remove deprecated API and properties

### LINKED ISSUE #172: mavenLocal() usage conflicts with globally-configured repositories in Gradle 6.8
Source of the issue https://github.com/vanniktech/gradle-maven-publish-plugin/blob/1f3ed6de7235a1ffb53e74c392c5d280bd3381fb/src/main/kotlin/com/vanniktech/maven/publish/MavenPublishPluginExtension.kt#L23-L27

see https://github.com/gradle/gradle/issues/15336#issuecomment-734711294

gabrielittner:
Removes the deprecated target API and throws an exception when running into deprecated properties. I've left the deprecated `uploadArchives` and `installArchives` tasks because there is no cost in supporting them.

~With this there is no repository at all configured, I will add configuring maven central in a separate PR with the new base API. We can wait with merging this one until the other one is ready.~

Fixes #172 (we don't call that method anymore).


src/main/kotlin/com/vanniktech/maven/publish/MavenPublishPlugin.kt
```diff
@@ -26,7 +28,20 @@ open class MavenPublishPlugin : Plugin<Project> {
...
+      repo.name = "mavenCentral"
+      repo.setUrl("https://oss.sonatype.org/service/local/staging/deploy/maven2/")
+      repo.credentials(PasswordCredentials::class.java)
+
+      p.afterEvaluate {
+        if (it.version.toString().endsWith("SNAPSHOT")) {
+          repo.setUrl("https://oss.sonatype.org/content/repositories/snapshots/")
+        }
+      }
+    }
```
gabrielittner, line 44: This will move again with the new API. Before the repo was just called maven but I think it's more clear if Gradle's tasks are called `publish...ToMavenCentralRepository` and not `publish...ToMavenRepository`

ZacSweers (approved):

vanniktech (approved):

src/main/kotlin/com/vanniktech/maven/publish/MavenPublishPlugin.kt
```diff
@@ -26,7 +28,20 @@ open class MavenPublishPlugin : Plugin<Project> {
 
     val pom = MavenPublishPom.fromProject(p)
     p.setCoordinates(pom)
-    p.configureTargets(extension)
+    p.checkProperties()
+    p.configureArchivesTasks()
+
+    p.publishing.repositories.maven { repo ->
+      repo.name = "mavenCentral"
```
martinbonnin, line 35: Nitpicking: the repository itself is OSS staging. But everyone knows about MavenCentral so I guess it makes sense. I'm not sure which one is best.

gabrielittner: I went with that for easier understandability. When using it, that's where you want your artifacts to end up with and most probably don't know all the details https://twitter.com/saketme/status/1357436828861759491

PR #195, opened 2021-02-14
fix artifact id in Kotlin MPP projects

### LINKED ISSUE #191: Changes in the pushed module names?
After the recent fixes for multiplatform modules in Kotlin 1.4.30 that went into version 0.14.0, I noticed a change in the structure and naming of the https://github.com/kirill-grouchnikov/aurora modules that were uploaded to Sonatype staging.

https://github.com/kirill-grouchnikov/aurora/tree/icicle/icon/icon used to go to https://oss.sonatype.org/content/repositories/snapshots/org/pushing-pixels/aurora-icon-icon/ but now is in https://oss.sonatype.org/content/repositories/snapshots/org/pushing-pixels/aurora-aurora-icon-icon-aurora-icon-icon/

Similar "changes" happened to the other Aurora modules. And what used to be

```
api("org.pushing-pixels:aurora-window:0.0.4-SNAPSHOT")
```

in the `dependencies` block of a sample app, now has to be

```
api("org.pushing-pixels:aurora-aurora-window:0.0.8-SNAPSHOT")
```

Is this intended?

gabrielittner:
Fixes #191.

In #180 I changed the artifact id logic to not run in `afterEvaluate`. However mpp projects write over our artifact id again so for those I kept the `afterEvaluate`. The issue is that for some of the publication our artifact id survives until afterEvaluate. The common and the metadata publications have the intended id while jvm/android/js/native are back to the original. If the new artifact id contains the `project.name` then we run into #191.  
Let's say the project name is `foo` and our artifact id is `foo-bar`. Initially the ids are `foo`, `foo-metadata` and `foo-jvm`. We update that to `foo-bar`, `foo-bar-metadata` and `foo-bar-jvm`. Then Kotlin changes it to `foo-bar`, `foo-bar-metadata` and `foo-jvm`. Now in after evaluate we replace `foo` with `foo-bar` again which is why we get `foo-bar-bar`, `foo-bar-bar-metadata` and `foo-bar-jvm`.

To fix this the afterEvaluate is now per publication and only changes the artifact id if it is not the one we set initially.

This is targeting `0.14.2-release` to do another bugfix release. I will merge that branch into master after the release.

### DIFF
--- a/src/integrationTest/kotlin/com/vanniktech/maven/publish/MavenPublishPluginIntegrationTest.kt
+++ b/src/integrationTest/kotlin/com/vanniktech/maven/publish/MavenPublishPluginIntegrationTest.kt
@@ -179,6 +179,38 @@ class MavenPublishPluginIntegrationTest(
     assertPomContentMatches(linuxArtifactId)
   }
 
+  @Test fun kotlinMppArtifactIdReplacementWorksCorrectly1() {
+    setupFixture("passing_kotlin_mpp_project", "foo")
+
+    val result = executeGradleCommands(uploadArchivesTargetTaskName, "--info", "--stacktrace", "-PPOM_ARTIFACT_ID=foo-bar")
+
+    assertThat(result.task(":$uploadArchivesTargetTaskName")?.outcome).isEqualTo(SUCCESS)
+    assertThat(result.task(":dokka")?.outcome).isEqualTo(SUCCESS)
+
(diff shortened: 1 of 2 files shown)


vanniktech (approved):

gabrielittner:
I believe the workflow failure for pull request is coming from initially having master as base branch.

PR #199, opened 2021-02-20
add base plugin

### LINKED ISSUE #171: Eliminate property file usage

gabrielittner:
This starts the implementation of the new APIs from #173.

- add a new `com.vanniktech.maven.publish.base` plugin
  - applies Gradle's `maven-publish`
  - makes the Gradle version check
  - creates the base extension
- adds a new extension
  - It's called `mavenPublishing`, because we already have `mavenPublish` in the old/current plugin. I eventually see the old extension going away. The name is based on the default `publishing` extension name from Gradle.
  - marked as incubating for now
  - has an API for configuring the POM, fixes #171 
  - other APIs will be added later on

The old/current plugin was updated to apply the base plugin. It will also apply the Gradle property POM values using the new API from base extension, instead of duplicating that logic.


src/main/kotlin/com/vanniktech/maven/publish/legacy/Pom.kt
```diff
@@ -0,0 +1,39 @@
+package com.vanniktech.maven.publish.legacy
+
+import com.vanniktech.maven.publish.baseExtension
+import com.vanniktech.maven.publish.findOptionalProperty
+import org.gradle.api.Project
+
+internal fun Project.configurePom() {
+  // without afterEvaluate https://github.com/gradle/gradle/issues/12259 will happen
+  afterEvaluate {
```
gabrielittner, line 9: That Gradle bug is kind of terrible for us. It means anyone using the new POM API on a Gradle plugin will need an afterEvaluate.

gabrielittner: After writing this I realised that it probably makes sense to change the code below to to null checks before calling set to avoid writing over explicit API calls with null

src/main/kotlin/com/vanniktech/maven/publish/MavenPublishBaseExtension.kt
```diff
@@ -0,0 +1,26 @@
+package com.vanniktech.maven.publish
+
+import org.gradle.api.Action
+import org.gradle.api.Incubating
+import org.gradle.api.Project
+import org.gradle.api.publish.maven.MavenPom
+import org.gradle.api.publish.maven.MavenPublication
+
+@Incubating
+open class MavenPublishBaseExtension(
```
ZacSweers, line 10: you can make this `abstract` instead

src/main/kotlin/com/vanniktech/maven/publish/MavenPublishBasePlugin.kt
```diff
@@ -0,0 +1,26 @@
...
+import org.gradle.api.Project
+import org.gradle.util.VersionNumber
+import org.gradle.api.publish.maven.plugins.MavenPublishPlugin as GradleMavenPublishPlugin
+
+open class MavenPublishBasePlugin : Plugin<Project> {
+
+  override fun apply(project: Project) {
+    val gradleVersion = VersionNumber.parse(project.gradle.gradleVersion)
+    if (gradleVersion < VersionNumber(MINIMUM_GRADLE_MAJOR, MINIMUM_GRADLE_MINOR, MINIMUM_GRADLE_MICRO, null)) {
+      throw IllegalArgumentException("You need Gradle version 6.6.0 or higher")
```
ZacSweers, line 13: Wouldn't this be more of a state thing anyway?

```suggestion
      error("You need Gradle version 6.6.0 or higher")
```

src/main/kotlin/com/vanniktech/maven/publish/MavenPublishBasePlugin.kt
```diff
@@ -0,0 +1,26 @@
...
+
+    project.plugins.apply(GradleMavenPublishPlugin::class.java)
+
+    project.extensions.create("mavenPublishing", MavenPublishBaseExtension::class.java, project)
+  }
+
+  private companion object {
+    const val MINIMUM_GRADLE_MAJOR = 6
+    const val MINIMUM_GRADLE_MINOR = 6
+    const val MINIMUM_GRADLE_MICRO = 0
```
ZacSweers, line 24: What about just saving a `val MIN_GRADLE_VERSION = VersionNumber.parse("6.6.0")`?

ZacSweers (approved):

src/main/kotlin/com/vanniktech/maven/publish/MavenPublishBaseExtension.kt
gabrielittner: 328210f

src/main/kotlin/com/vanniktech/maven/publish/MavenPublishBasePlugin.kt
gabrielittner: 6a38ada

src/main/kotlin/com/vanniktech/maven/publish/MavenPublishBasePlugin.kt
gabrielittner: 👍 6a38ada

PR #200, opened 2021-02-21
add publishToMavenCentral API

gabrielittner:
Part of #173 

Adds `publishToMavenCentral()` to the base extension which will:
- add a repository called `mavenCentral`
- configure the url based on version
- use the credentials API for username/password
- set up `NexusOptions` using the same properties that come from the credentials API

Other changes:
- Main plugin uses the new API to add the default repository.
- Moved adding the `closeAndReleaseRepository` task to the base plugin and only do it if the nexus API was used. For users of the main plugin this isn't a behavior change because it's automatically called through `publishToMavenCentral()` for them.

Omissions from the API:
- There is no `stagingProfile` parameter on `publishToMavenCentral`. If there is just one open staging repository we will pick that, even if `stagingProfile` is not set. For cases where you might have multiple open at the same time from different project it's not unlikely that they use the same profile, so it wouldn't help to resolve conflicts in many cases. The other reason is that if we add something to either explicitly upload to a specific staging repository, we'd need the full staging repository id. If we'd go a step further and would add a task to create a new staging repository then we'd need the id of the staging profile and not the name that we have right now (https://support.sonatype.com/hc/en-us/articles/213465868-Uploading-to-a-Staging-Repository-via-REST-API)
- `nexusOptions` is internal in the base plugin for now. I'm not sure if staging repositories are used outside of publishing to maven central and if companies with their own Nexus instances would just publish directly to their repositories. That's why I would keep it out for now until we get a feature request for it.

As for signing, it's not automatically set up by `publishToMavenCentral()`. If you're using SNAPSHOT versions for regular builds, then signing would still not be required, but we had requests in the pasts for the ability to create unsigned non snapshot builds. That's why the old extension has a `releaseSigningEnabled` option. So instead of adding an optional boolean parameter here, I'd just say it's not automatically enabled and you need to call `signAllPublications()` (#201) yourself in the base plugin.


martinbonnin:
Nice! A few comments:

> The other reason is that if we add something to either explicitly upload to a specific staging repository, we'd need the full staging repository id. 

A task to create a new staging repository would be nice. Even without this, there are  use cases where the repository id would be created outside the build and provided as an environment variable (see https://github.com/martinbonnin/create-nexus-staging-repo-sample). Especially for multiplatform builds, having all different VMs/job publish to the same predictable repository would remove some friction. So maybe an overload like:

```
publishToMavenCentral(repositoryId: String)
```

>  If we'd go a step further and would add a task to create a new staging repository then we'd need the id of the staging profile and not the name that we have right now

There's a `GET staging/profiles` endpoint on Nexus that lists all staging profiles with their id and name so the id could be looked up from the name if needed (and vice versa). Overall, I think most users will have only one staging profile in which case the first one can be used automatically. If multiple are found, we can throw a nice actionable error with the different options.

> nexusOptions is internal in the base plugin for now. [...] That's why I would keep it out for now until we get a feature request for it. 

Makes sense.

gabrielittner:
The overload makes sense 👍 How would you imagine it's interaction with snapshots? Should it be nullable and we fail when the version is a SNAPSHOT and the id is not null? Or would you call the no arg version for non snapshots which would be more verbose on the caller side.

martinbonnin:
> How would you imagine it's interaction with snapshots?

Good point about SNAPSHOTs. 

I would certainly have a separate `OSSSnapshots` repo be configured all the time and the decision whether to publish to snapshots or staging be left to the ci scripts:
* if(main) ./gradlew publishAllPublicationsToSnapshotsRepository
* if (tag) ./gradlew publishAllPublicationsToMavenCentralRepository

So maybe this calls for 2 additional APIs:

```kotlin
// Simple, batteries-included API
// publishes to Snapshots if version has -SNAPSHOT or to an implicit staging repo else
publishToMavenLocal()

// Advanced API for users that want more granular control over the repositories
// creates the "OSSStaging" repository
publishToOSSStagingRepository(stagingRepositoryId: String)
// creates the "OSSSnapshot" repository
publishToOSSSnapshotsRepository()
```

eygraber:
You might want to address the fact that the sonatype nexus url being used is now "legacy" - https://github.com/vanniktech/gradle-maven-publish-plugin/issues/206

gabrielittner:
> // Simple, batteries-included API
// publishes to Snapshots if version has -SNAPSHOT or to an implicit staging repo else
publishToMavenLocal()

Gradle already has something built in for this. There is a `publishToMavenLocal` task by default.

> // Advanced API for users that want more granular control over the repositories
// creates the "OSSStaging" repository
publishToOSSStagingRepository(stagingRepositoryId: String)
// creates the "OSSSnapshot" repository
publishToOSSSnapshotsRepository()

My issue with having a separate `publishToOSSSnapshotsRepository()` repository  is that then either users of the plugin need to do the logic of `if (version.endsWith("SNAPSHOT") publishToOSSSnapshotsRepository() else publishToMavenCentral()` which is something we're handling right now or they just add both and always have an invalid repository. With the `s01.oss.sonatype` stuff this becomes more complex. I think we could just have: `publishToMavenCentral(optionForS01: ProbablyAnEnum, stagingRepositoryId: String? = null)`. Then if you don't pass in a `stagingRepositoryId` it behaves like right now and dynamically chooses the right url based on the version. If you pass in a `stagingRepositoryId` and the current version is a `SNAPSHOT` we fail (if we'd do nothing then Sonatype would fail when trying to upload).

martinbonnin:
> > // Simple, batteries-included API
> > // publishes to Snapshots if version has -SNAPSHOT or to an implicit staging repo else
> > publishToMavenLocal()
> 
> Gradle already has something built in for this. There is a `publishToMavenLocal` task by default.

Gaaa, my brain bugged, I meant `publishToMavenCentral()` as in the existing API 🤦 . Sorry, I edited that...

> My issue with having a separate `publishToOSSSnapshotsRepository()` repository is that then either users of the plugin need to do the logic of `if (version.endsWith("SNAPSHOT") publishToOSSSnapshotsRepository() else publishToMavenCentral()` which is something we're handling right now or they just add both and always have an invalid repository. 

Yep, fair

> With the `s01.oss.sonatype` stuff this becomes more complex.

Yup...

> I think we could just have: `publishToMavenCentral(optionForS01: ProbablyAnEnum, stagingRepositoryId: String? = null)`. Then if you don't pass in a `stagingRepositoryId` it behaves like right now and dynamically chooses the right url based on the version. If you pass in a `stagingRepositoryId` and the current version is a `SNAPSHOT` we fail (if we'd do nothing then Sonatype would fail when trying to upload).

Makes sense and it's more consistent with the existing API. LGTM 👍

src/main/kotlin/com/vanniktech/maven/publish/legacy/BaseSetup.kt
```diff
@@ -1,9 +1,14 @@
 package com.vanniktech.maven.publish.legacy
 
+import com.vanniktech.maven.publish.SonatypeHost
 import com.vanniktech.maven.publish.baseExtension
 import com.vanniktech.maven.publish.findOptionalProperty
 import org.gradle.api.Project
 
+internal fun Project.configureMavenCentral() {
+  baseExtension.publishToMavenCentral(SonatypeHost.DEFAULT)
```
gabrielittner, line 9: I'll add something for this to the legacy extension in a separate PR.

gabrielittner:
Updated with parameters for the new Sonatype host and for staging repository id.

src/main/kotlin/com/vanniktech/maven/publish/MavenPublishBaseExtension.kt
```diff
@@ -12,6 +16,71 @@ abstract class MavenPublishBaseExtension(
...
+      it.baseUrl = "${host.rootUrl}/service/local/"
+      it.repositoryUsername = project.findOptionalProperty("mavenCentralRepositoryUsername")
+      it.repositoryPassword = project.findOptionalProperty("mavenCentralRepositoryPassword")
+    }
+  }
+
+  internal fun nexusOptions(action: Action<NexusOptions>) {
+    var nexusOptions = this.nexusOptions
+    if (nexusOptions == null) {
+      nexusOptions = NexusOptions(null, null, null, null)
```
ZacSweers, line 66: should we be using `objects.newInstance<NexusOptions>()` to be more gradle-friendly here?

src/main/kotlin/com/vanniktech/maven/publish/MavenPublishBaseExtension.kt
```diff
@@ -12,6 +16,71 @@ abstract class MavenPublishBaseExtension(
...
+      this.nexusOptions = nexusOptions
+
+      @Suppress("SwallowedException")
+      try {
+        project.rootProject.tasks.named("closeAndReleaseRepository")
+      } catch (e: UnknownTaskException) {
+        project.rootProject.tasks.register("closeAndReleaseRepository", CloseAndReleaseRepositoryTask::class.java) {
```
ZacSweers, line 73: I would just make the task regardless, worst case it's unused. Tasks should be free

src/main/kotlin/com/vanniktech/maven/publish/MavenPublishBaseExtension.kt
gabrielittner: 704f143

src/main/kotlin/com/vanniktech/maven/publish/MavenPublishBaseExtension.kt
gabrielittner: aba6546

gabrielittner: Reverted in df3ad06 because of a misunderstanding. Is it ok if I replace this part in a separate PR, because we used the same approach in the old code as well?

ZacSweers: 👍

PR #209, opened 2021-03-21
workaround AndroidJavadoc task failing on newer jdks

gabrielittner:
This makes it possible to use the plugin on 9+. I couldn't add 14 to Github Actions because jacoco fails then, will look at that separately


ZacSweers (approved):

vanniktech (commented):
👍

PR #212, opened 2021-03-28
add API to configure what to publish

gabrielittner:
This adds the API tell us what should be published, e.g. an Android library with sources jars and empty javadoc jars. The idea is still the same as in the design doc but instead of different plugins it's all in the base plugin. I'm not 100% if this is the best approach from the perspective of how Gradle APIs usually look like, but I think it can work. The main plugin uses these APIs to configure what to publish as well.

2 behavior changes:
- Dokka is not forced anymore on jvm and Kotlin android projects, the main plugin will use it's output when it's there, otherwise it will go for regular or empty javadoc jars
- removed publishing an extra groovydoc jar when the groovy plugin is present

~There are still some TODOs for dokka in the integration tests and some docs, but I wanted to already open this to get feedback.~


src/main/kotlin/com/vanniktech/maven/publish/MavenPublishBaseExtension.kt
```diff
@@ -136,4 +151,32 @@ abstract class MavenPublishBaseExtension(
       it.pom(configure)
     }
   }
+
+  /**
+   * Configures a [Platform] which will automatically set up the artifacts that should get published, including javadoc
+   * and sources jars depending on the option.
+   */
+  @Incubating
+  fun configure(platform: Platform) {
```
gabrielittner, line 160: This might not actually work well in Groovy. Any ideas on alternatives?

src/main/kotlin/com/vanniktech/maven/publish/MavenPublishBaseExtension.kt
```diff
@@ -136,4 +151,32 @@ abstract class MavenPublishBaseExtension(
...
+    if (this.platform != null) {
+      // ignore subsequent calls with the same arguments
```
vanniktech, line 162:
```suggestion
      // Ignore subsequent calls with the same arguments.
```

src/main/kotlin/com/vanniktech/maven/publish/Platform.kt
```diff
@@ -0,0 +1,165 @@
...
+ *
+ * java {
+ *   withSourcesJar()
+ *   withJavadocJar()
+ * }
+ ```
+ */
+data class JavaLibrary(
+  override val javadocJar: JavadocJar,
+  override val sourcesJar: Boolean = true
```
vanniktech, line 37:
```suggestion
  override val sourcesJar: Boolean = true,
```

src/main/kotlin/com/vanniktech/maven/publish/Platform.kt
```diff
@@ -0,0 +1,165 @@
...
+ * ```
+ * java {
+ *   withSourcesJar()
+ *   withJavadocJar()
+ * }
+```
+ */
+data class GradlePlugin(
+  override val javadocJar: JavadocJar,
+  override val sourcesJar: Boolean = true
```
vanniktech, line 55:
```suggestion
  override val sourcesJar: Boolean = true,
```

src/main/kotlin/com/vanniktech/maven/publish/Platform.kt
```diff
@@ -0,0 +1,165 @@
...
+ *     }
+ *   }
+ * }
+ * ```
+ * This does not include javadoc and sources jars because there are no APIs for that available.
+ */
+data class AndroidLibrary(
+  override val javadocJar: JavadocJar,
+  override val sourcesJar: Boolean = true,
+  val variant: String = "release"
```
vanniktech, line 80:
```suggestion
  val variant: String = "release",
```

src/main/kotlin/com/vanniktech/maven/publish/Platform.kt
```diff
@@ -0,0 +1,165 @@
...
+ * `-javadoc` will be added to the publications.
+ *
+ * Equivalent Gradle set up:
+ * `n/a`
+ * This does not include javadoc jars because there are no APIs for that available.
+ */
+data class KotlinMultiplatform(
+  override val javadocJar: JavadocJar = JavadocJar.Empty
+) : Platform() {
+  // automatically added by Kotlin MPP plugin
```
vanniktech, line 95:
```suggestion
  // Automatically added by Kotlin MPP plugin.
```

src/main/kotlin/com/vanniktech/maven/publish/Platform.kt
```diff
@@ -0,0 +1,165 @@
...
+ * To be used for `org.jetbrains.kotlin.multiplatform` projects.  Uses the default publications that gets created by
+ * that plugin, including the automatically created `-sources` jars. Depending on the passed parameters for [javadocJar],
+ * `-javadoc` will be added to the publications.
+ *
+ * Equivalent Gradle set up:
+ * `n/a`
+ * This does not include javadoc jars because there are no APIs for that available.
+ */
+data class KotlinMultiplatform(
+  override val javadocJar: JavadocJar = JavadocJar.Empty
```
vanniktech, line 93:
```suggestion
  override val javadocJar: JavadocJar = JavadocJar.Empty,
```

src/main/kotlin/com/vanniktech/maven/publish/Platform.kt
```diff
@@ -0,0 +1,165 @@
...
+ * This does not include javadoc and sources jars because there are no APIs for that available.
+ */
+data class AndroidLibrary(
+  override val javadocJar: JavadocJar,
+  override val sourcesJar: Boolean = true,
+  val variant: String = "release"
+) : Platform()
+
+/**
+ * To be used for `org.jetbrains.kotlin.multiplatform` projects.  Uses the default publications that gets created by
```
vanniktech, line 84:
```suggestion
 * To be used for `org.jetbrains.kotlin.multiplatform` projects. Uses the default publications that gets created by
```

src/main/kotlin/com/vanniktech/maven/publish/Platform.kt
```diff
@@ -0,0 +1,165 @@
...
+ *     from(components["java"])
+ *     artifact(project.tasks.named("kotlinSourcesJar"))
+ *   }
+ * }
+ * ```
+ * This does not include javadoc jars because there are no APIs for that available.
+  */
+data class KotlinJvm(
+  override val javadocJar: JavadocJar = JavadocJar.Empty,
+  override val sourcesJar: Boolean = true
```
vanniktech, line 117:
```suggestion
  override val sourcesJar: Boolean = true,
```

src/main/kotlin/com/vanniktech/maven/publish/Platform.kt
```diff
@@ -0,0 +1,165 @@
...
+ *     from(components["kotlin"])
+ *     artifact(project.tasks.named("kotlinSourcesJar"))
+ *   }
+ * }
+ * ```
+ * This does not include javadoc jars because there are no APIs for that available.
+ */
+data class KotlinJs(
+  override val javadocJar: JavadocJar = JavadocJar.Empty,
+  override val sourcesJar: Boolean = true
```
vanniktech, line 138:
```suggestion
  override val sourcesJar: Boolean = true,
```

src/main/kotlin/com/vanniktech/maven/publish/legacy/BaseSetup.kt
```diff
@@ -78,3 +89,58 @@ internal fun Project.configurePom() {
     }
   }
 }
+
+internal fun Project.configurePlatform() {
+  afterEvaluate {
+
+    when {
```
vanniktech, line 96: Nice!

vanniktech (approved):
Just nitpicking. Rest looks good 👍

gabrielittner:
I've left the trailing coma suggestions. We're still on an older Kotlin version because some Gradle stuff breaks when updating it, still need to look into that

PR #227, opened 2021-04-24
update README and CHANGELOG for 0.15.0

gabrielittner:


README.md
```diff
@@ -15,16 +15,12 @@ buildscript {
...
     classpath 'com.vanniktech:gradle-maven-publish-plugin:0.14.2'
-    // For Kotlin projects, you need to add Dokka.
-    classpath 'org.jetbrains.dokka:dokka-gradle-plugin:1.x.x'
   }
 }
 
 apply plugin: "com.vanniktech.maven.publish"
 ```
 
-Information: [This plugin is also available on Gradle plugins](https://plugins.gradle.org/plugin/com.vanniktech.maven.publish)
```
gabrielittner, line 26: Removed this because our automation does not publish to this.

README.md
```diff
@@ -33,29 +29,24 @@ To configure the coordinates of your published artifacts as well as the POM this
...
 POM_LICENCE_URL=https://www.apache.org/licenses/LICENSE-2.0.txt
 POM_LICENCE_DIST=repo
 
+POM_SCM_URL=https://github.com/username/mylibrary/
+POM_SCM_CONNECTION=scm:git:git://github.com/username/mylibrary.git
+POM_SCM_DEV_CONNECTION=scm:git:ssh://git@github.com/username/mylibrary.git
+
 POM_DEVELOPER_ID=username
 POM_DEVELOPER_NAME=User Name
 POM_DEVELOPER_URL=https://github.com/username/
```
gabrielittner, line 52: Made this into one block because all of these are required for maven central

CHANGELOG.md
```diff
@@ -1,8 +1,17 @@
...
+Version 0.15.0 *(2021-04-24)*
 ---------------------------------
 
+- **BREAKING**: Removed support for deprecated `RELEASE_REPOSITORY_URL`, `SNAPSHOT_REPOSITORY_URL`, `SONATYPE_NEXUS_USERNAME`, `SONATYPE_NEXUS_PASSWORD` environment variables and properties. 
+  For safety reasons the project will fail when finding these. Use `mavenCentralRepositoryUsername` and `mavenCentralRepositoryPassword` Gradle properties or 
+  `ORG_GRADLE_PROJECT_mavenCentralRepositoryUsername` and `ORG_GRADLE_PROJECT_mavenCentralRepositoryPassword` environment variables instead.
+- **BREAKING**: Removed deprecated `targets` API. See README for alternative ways of adding targets.
+- Support for `s01.oss.sonatype.org` by setting `sonatypeHost = "S01"`.
+- Introduce `com.vanniktech.maven.publish.base` plugin. This plugin contains all the functionality of the main plugin, but does not configure anything automatically. 
+  Instead, it offers a public API which is also used by the main plugin to do so yourself. This allows for more flexibility and to publish project types.
```
vanniktech, line 12:
```suggestion
  Instead, it offers a public API, which is also used by the main plugin to do so yourself. This allows for more flexibility and to publish different project types. 
```

README.md
vanniktech: Right. I used to publish this by hand

README.md
```diff
@@ -112,6 +115,17 @@ signing.secretKeyRingFile=/Users/yourusername/.gnupg/secring.gpg
 It's best to place them inside your home directory, `$HOME/.gradle/gradle.properties`. You can find more information
 about these properties in [Gradle's documentaion](https://docs.gradle.org/current/userguide/signing_plugin.html#sec:signatory_credentials).
 
+In case you want to use in memory signing keys, which works great for CI, you can specify them like this instead:
+```properties
+signingInMemoryKey=exported_ascii_armored_key
+# optional
```
vanniktech, line 121:
```suggestion
# Optional.
```

README.md
```diff
@@ -112,6 +115,17 @@ signing.secretKeyRingFile=/Users/yourusername/.gnupg/secring.gpg
...
+signingInMemoryKeyId=24875D73
+# if key was created with a password
```
vanniktech, line 123:
```suggestion
# If key was created with a password.
```

README.md
```diff
@@ -142,25 +156,7 @@ mavenPublish {
...
-    }
-}
-```
-The `stagingProfile` set here is either the same as your group id or a simpler version of it. When you are publishing a
-library with `com.example.mylibrary` as group then it would either be the same or just `com.example`. You can find it
-by looking at your [Sonatype staging profiles](https://oss.sonatype.org/#stagingProfiles) in the name and repo target
-columns.
-
-This will create a `closeAndReleaseRepository` task that you can call after `publish`:
+Alternatively, the plugin will create a `closeAndReleaseRepository` task that you can call after `publish`:
```
vanniktech, line 159:
```suggestion
Additionally, the plugin will create a `closeAndReleaseRepository` task that you can call after `publish`:
```

README.md
```diff
@@ -170,6 +166,90 @@ This will create a `closeAndReleaseRepository` task that you can call after `pub
...
+```groovy
+import com.vanniktech.maven.publish.SonatypeHost
+
+allprojects {
+    plugins.withId("com.vanniktech.maven.publish.base") {
+        GROUP = "com.example.project"
+        VERSION = "1.0.3-SNAPSHOT"
+        
+        mavenPublishing {
+            // alternatively "S01"
```
vanniktech, line 186: Let's remove it. It's mentioned above

```suggestion

```

README.md
```diff
@@ -170,6 +166,90 @@ This will create a `closeAndReleaseRepository` task that you can call after `pub
...
+            publishToMavenCentral("DEFAULT")
```
vanniktech, line 187: Can we get?

```suggestion
            publishToMavenCentral()
```

README.md
```diff
@@ -170,6 +166,90 @@ This will create a `closeAndReleaseRepository` task that you can call after `pub
...
+            
+            // will only apply to non snapshot builds
```
vanniktech, line 189:
```suggestion
            // Will only apply to non snapshot builds.
```

README.md
```diff
@@ -170,6 +166,90 @@ This will create a `closeAndReleaseRepository` task that you can call after `pub
...
+            // uses credentials as described above, supports both regular and in memory signing
```
vanniktech, line 190:
```suggestion
            // Uses credentials as described above, supports both regular and in memory signing.
```

vanniktech (approved):
Thanks for all of your work @gabrielittner

README.md
gabrielittner: It doesn't have a default value right now. The reason why I didn't in the base plugin API is that since you have to call this anyways I though it's nicer to have it explicit. More and more users will have s01. No strong opinion though

PR #230, opened 2021-04-28
Use mavenCentralUsername and mavenCentralPassword everywhere

### LINKED ISSUE #229: MavenCentral username and password properties no longer work
The README says to use `mavenCentralRepositoryUsername` and `mavenCentralRepositoryPassword` but that fails when running the publish task. It says that `mavenCentralUsername` and `mavenCentralPassword` are missing. Renaming the properties to drop `Repository` works.

eygraber:
Fixes #229 

Not sure if this is what you had in mind, but figured I'd give it a shot.

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,9 +78,9 @@ allprojects {
 }
 ```
 
-The username and password for Sonatype OSS can be provided as Gradle properties called `mavenCentralRepositoryUsername` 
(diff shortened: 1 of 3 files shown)


gabrielittner (approved):
👍 

Yes that's exactly it. The credentials for uploading are configured like this https://docs.gradle.org/current/samples/sample_publishing_credentials.html For some reason I've kept the `Repository` part of the sample when I added the properties for nexusOptions

PR #231, opened 2021-04-28
Fallback to an empty string when signingInMemoryKeyPassword is null

eygraber:


gabrielittner (approved):
Nice, thanks for adding this

PR #239, opened 2021-05-03
update dependencies

gabrielittner:
Updated most dependencies and removed some obsolete plugins. I've kept dokka where it is because it runs out of memory on CI often when using newer versions. AGP is kept on an older release because that's the minimum we need.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,11 +1,11 @@
 buildscript {
-  ext.kotlinVersion = '1.4.30'
-  ext.junitVersion = '4.13'
-  ext.assertjVersion = '3.15.0'
+  ext.kotlinVersion = '1.4.32'
+  ext.junitVersion = '4.13.2'
+  ext.assertjVersion = '3.19.0'
   ext.agpVersion = '3.6.0'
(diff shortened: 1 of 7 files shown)


ZacSweers (approved):

build.gradle
```diff
@@ -29,17 +27,15 @@ apply plugin: 'kotlin'
...
-apply plugin: 'com.gradle.plugin-publish'
 apply from: "gradle/integration-test.gradle"
 
 codeQualityTools {
   ktlint {
-    toolVersion = '0.31.0'
+    toolVersion = '0.40.0'
   }
   detekt {
-    toolVersion = '1.0.0'
```
vanniktech, line 42: Feel free to remove. I've stopped using detekt

vanniktech (approved):

build.gradle
gabrielittner: Will do that in a separate pr 👍

gabrielittner:
Downgraded Gradle for now, I will update it again separately updating the integration tests that fail

PR #244, opened 2021-05-07
provide an option to not publish the default repository

jiqimaogou:
the plugin add a default maven repository even I add the repository according to the doc:

You can add additional repositories to publish to using the standard Gradle APIs:

```groovy
publishing {
    repositories {
        maven {
            def releasesRepoUrl = "$buildDir/repos/releases"
            def snapshotsRepoUrl = "$buildDir/repos/snapshots"
            url = version.endsWith('SNAPSHOT') ? snapshotsRepoUrl : releasesRepoUrl
        }
    }
}
```

In company, the library may never publish to public repository. if I execute `publish` task, it  will publish the default maven repository too.

### DIFF
--- a/src/main/kotlin/com/vanniktech/maven/publish/MavenPublishPluginExtension.kt
+++ b/src/main/kotlin/com/vanniktech/maven/publish/MavenPublishPluginExtension.kt
@@ -21,7 +21,7 @@ abstract class MavenPublishPluginExtension(
    *
    * @Since 0.15.0
    */
-  var sonatypeHost: SonatypeHost = SonatypeHost.DEFAULT
+  var sonatypeHost: SonatypeHost? = SonatypeHost.DEFAULT
 
   /**
    * The Android library variant that should be published. Projects not using any product flavors, that just want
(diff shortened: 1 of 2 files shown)


gabrielittner:
I liked your initial implementation where you made sonatypeHost optional more. It is very explicit and does not rely on ordering of calls or naming of the repository. What was the reason for changing it?

jiqimaogou:
> I liked your initial implementation where you made sonatypeHost optional more. It is very explicit and does not rely on ordering of calls or naming of the repository. What was the reason for changing it?

If I do not want the publication to the default repository. I must set `sonatypeHost` to `null` explicitly. the new implementation make this implicitly.

if you like the initial implementation, I will revert to initial implementation.

gabrielittner:
I'd go with the initial one, it's more predictable behavior wise. We'll also have a Gradle property to disable it though (#245)

jiqimaogou:
> I'd go with the initial one, it's more predictable behavior wise. We'll also have a Gradle property to disable it though (#245)

has do it.

gabrielittner:
Thank you for the pr 👍

PR #250, opened 2021-05-14
add pomFromGradleProperties API to base plugin

gabrielittner:
I think it's fine to have both the property and the DSL based pom configuration going forward, which means that this functionality should live in the base plugin. Note that this does not included project group, version and artifact id.


src/main/kotlin/com/vanniktech/maven/publish/legacy/BaseSetup.kt
```diff
@@ -36,67 +35,10 @@ internal fun Project.configureSigning() {
...
-          }
-        }
-      }
-    }
-  }
+  baseExtension.pomFromGradleProperties()
 }
 
-@Suppress("ComplexMethod")
+@Suppress("ComplexMethod", "LongMethod")
```
vanniktech, line 41: Let's just kill detekt ...

vanniktech (approved):

PR #261, opened 2021-05-26
Support configuration of androidJavaDocs task not to add reference links

panpf:
Because the androidJavaDocs task requires access to https://developer.android.com/reference, and China has blocked android.com, it often fails when used in China


panpf:
Please approve the merger!

Issue #248, opened 2021-05-14
Cannot sign with in-memory key '[...]:signMavenPublication' because it has no configured signatory


victorpaleologue:
As explained in the README: https://github.com/vanniktech/gradle-maven-publish-plugin#signing
I moved to in-memory GPG keys, using the following Gradle properties:
```ini
signingInMemoryKey=...
signingInMemoryKeyId=...
signingInMemoryPassword=...
```
I used `gpg --export-secret-keys --armor $KEY_ID` and assigned the data to `signingInMemoryKey`.
I also tried with the base64 version.
It really seems the key is well decrypted, because when I did it wrong I had the error `Could not read PGP secret key`, and I do not have it here.

I did not change my gradle code, but now it fails with the message: `Cannot perform signing task '[...]:signMavenPublication' because it has no configured signatory`
You can see the gradle code here: https://github.com/aldebaran/pddl-planning-android/blob/a97d0656b7fe65cc0668b37c9e0da89e044932f0/pddl-planning/build.gradle

And the full error report with the stacktrace here: https://pastebin.com/3aiRjsZz

I am probably doing something wrong, but I cannot find any clue about it in the README.

victorpaleologue:
I am not sure I understand Gradle well, but here is the first serious clue I found.

Looking at the plug-in's code, the error message corresponds to the absence of a `Signatory` object. It seems that calling `useInMemoryPgpKeys` [is meant to set signatories](https://github.com/gradle/gradle/blob/5ec3f672ed600a86280be490395d70b7bc634862/subprojects/signing/src/main/java/org/gradle/plugins/signing/SigningExtension.java#L248).

In [Gradle's `signing` plug-in tests](https://github.com/gradle/gradle/blob/5ec3f672ed600a86280be490395d70b7bc634862/subprojects/signing/src/integTest/groovy/org/gradle/plugins/signing/InMemoryPgpSignatoryProviderIntegrationSpec.groovy), the only signing declaration to be tested puts the call to `useInMemoryPgpKeys` before the call to `sign`.

[...]

PR #266, opened 2021-06-06
Fix key password property name in documentation and publication action

twyatt:
According to [`MavenPublishBaseExtension.kt#L121`](https://github.com/vanniktech/gradle-maven-publish-plugin/blame/4b3df71b859bf5260d90542a12c991ed31d42af2/src/main/kotlin/com/vanniktech/maven/publish/MavenPublishBaseExtension.kt#L121), the Gradle property name for the in-memory signing key password is `signingInMemoryKeyPassword` (note the **Key** before "Password") whereas the documentation listed it without `Key` (i.e. as `signingInMemoryPassword`).

This documentation typo may also have been the source of confusion in #248?

I believe the `publish-release` GitHub Action is also using the wrong property name (and I suspect it would've failed on your next publication attempt?).


gabrielittner:
Thanks a lot for finding this and opening the pr.

> I believe the publish-release GitHub Action is also using the wrong property name (and I suspect it would've failed on your next publication attempt?).

Yes it would have failed.

PR #277, opened 2021-06-27
Convert build scripts to Kotlin

martinbonnin:
This is a 1:1 conversion, no functional change.

1st step towards making the project a multi-module project that can publish a separate nexus client artifact.

See https://github.com/vanniktech/gradle-maven-publish-plugin/issues/265

Issue #265, opened 2021-06-01
Nexus package to a separate artifact? 


yogurtearl:
Would you consider publishing everything in the nexus package as a separate artifact(s)? 

https://github.com/vanniktech/gradle-maven-publish-plugin/tree/master/src/main/kotlin/com/vanniktech/maven/publish/nexus

maybe `com.vannitech:nexus-staging-api` and `com.vannitech:gradle-nexus-close-release-task`? 

This would allow just pulling in the retrofit service and/or task.

martinbonnin:
+1 for that. Extracting the Nexus Client shouldn't be too complex. I think the most complex part would be transforming  the project in a multi-module build and handling the publishing, etc... Maybe move the scripts to `build.gradle.kts` ? I could look into that if we feel that's a route we want to take.

gabrielittner:
No objections against this from my side. Do you want to open a PR?

martinbonnin:
👍  Will look into this. It might be several PRs, the first one certainly moving the build scripts to Kotlin.

PR #279, opened 2021-07-05
Make the project a multi-module build

martinbonnin:
Follow up from https://github.com/vanniktech/gradle-maven-publish-plugin/pull/277 and step 2/3 to publish the nexus API as a separate artifact (see #265), make the project a multi module build.

I went the `buildSrc` route because I've had some success with it in the past but I'm happy to investigate other alternatives like versions catalog or other if we want to avoid the "build invalidation on version bump" issue.

Step 3/3 will be to actually create a `nexus` module to hold the nexus API that `plugin` will depend on.


gabrielittner (approved):
Sorry for taking so long. Looks good to me 👍 

> I went the buildSrc route because I've had some success with it in the past but I'm happy to investigate other alternatives like versions catalog or other if we want to avoid the "build invalidation on version bump" issue.

I think it's fine for small projects like this.

martinbonnin:
Whoops, my turn to miss the notification :/. I'll go back to this in the coming days and extract a `nexus` module.

PR #288, opened 2021-08-22
add property for SonatypeHost, move base calls directly to plugin

gabrielittner:
This adds a way to set `SONATYPE_HOST` as a Gradle property for consistency. 

I've also moved some of the logic of the main plugin back into that class, when it's just about calling the base extension properly. So the legacy package only contains actual legacy code.


vanniktech (approved):
Nice

PR #290, opened 2021-08-22
remove MavenPublishConfigurer

gabrielittner:
The MavenPublishConfigurer is a leftover from the times where we had to implementations, one for the old `maven` plugin and one for `maven-publish`. This moves the logic to a `configure` method in the individual platform classes. The implementation is now right next to the documentation and we safe the extra mapping step in the base extension.


vanniktech (approved):
👍

PR #295, opened 2021-08-31
Add .gitattributes

Goooler:
See [Add .gitattributes](https://github.com/square/okhttp/pull/6773).


vanniktech (commented):
Why would we need this here?

Goooler:
> Why would we need this here?

See [here](https://github.com/square/okhttp/pull/6773).

vanniktech (approved):

PR #300, opened 2021-09-02
Merge all actions workflows into build.yml

Goooler:
Run publish jobs depend on build.


gabrielittner:
Thank you for this, but I prefer keeping them in separate files.

PR #303, opened 2021-09-14
Optimize gradle

Goooler:
-  `gradle-versions-plugin` could be configured in init script for global, and should configured with rules, see [Using a Gradle init script](https://github.com/ben-manes/gradle-versions-plugin#using-a-gradle-init-script).
- Remove empty kts files.
- Generate new gradlew.


gabrielittner:
Thanks, but I prefer keeping the versions plugin inside the project

PR #321, opened 2021-12-29
add only aar or jar ex

yizems:


yizems:
Can provide an example ?

PR #327, opened 2022-01-29
change how integration tests are run, update dependencies/build

gabrielittner:
The tests now run on a locally published snapshot. This solves the issue that they either use AGP/Kotlin/Dokka from our compileOnly classpath, which means we can only test against one version, or that there are issues with finding classes. The setup is mostly copied from here https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin

I'm planning to refactor the integration tests completely after the next release to then actually run them against multiple Gradle, Kotlin and AGP versions and to simplify the fixture setup. For now this will just enable testing for the AGP 7.1 changes.

The old unit test was deleted as it didn't provide any value.

Also bumps all dependencies and Gradle to the latest version.


vanniktech (commented):
Well done!

vanniktech (approved):

PR #340, opened 2022-04-06
deprecate old mavenPublish extension

gabrielittner:
This deprecates the old `sonatypeHost` and `releaseSigningEnabled` properties on the DSL. It also prepares for the following release to not add maven central as a default publishing target and not to enable release signing by default anymore.

Everyone currently publishing through maven central or signing releases without explicitly configuring it will get a warning. To get rid of the warning and prepare for the new behavior 
```
SONATYPE_HOST=DEFAULT
RELEASE_SIGNING_ENABLED=true
```
(`SONATYPE_HOST=S01` if s01.sonatype.org is used) needs to be added to the Gradle properties. Alternatively it's possible to use the new extension:
```
mavenPublishing {
    publishToMavenCentral("DEFAULT")
    signAllPublications()
}
```
Those 2 methods are now also not incubating anymore.

I've already updated the README because the properties are already fully supported.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,9 @@ uses Gradle properties. It's generally recommended to set them in your `gradle.p
 file.
 
 ```properties
+SONATYPE_HOST=DEFAULT
+RELEASE_SIGNING_ENABLED=true
+
 GROUP=com.test.mylibrary
 POM_ARTIFACT_ID=mylibrary-runtime
 VERSION_NAME=3.0.5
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -30,6 +30,9 @@ uses Gradle properties. It's generally recommended to set them in your `gradle.p
 file.
 
 ```properties
+SONATYPE_HOST=DEFAULT
```
vanniktech, line 33: I really don't like that you have to specify this, why can't we do `default` always and only if you want to change it, you have to declare it

README.md
```diff
@@ -30,6 +30,9 @@ uses Gradle properties. It's generally recommended to set them in your `gradle.p
...
+RELEASE_SIGNING_ENABLED=true
```
vanniktech, line 34: Same here as above

plugin/src/main/kotlin/com/vanniktech/maven/publish/MavenPublishPluginExtension.kt
```diff
@@ -37,10 +38,19 @@ abstract class MavenPublishPluginExtension(
    *
    * @Since 0.9.0
    */
+  @Deprecated("Set the RELEASE_SIGNING_ENABLED Gradle property or call mavenPublishing { signAllPublications() } instead")
   var releaseSigningEnabled: Boolean = project.findOptionalProperty("RELEASE_SIGNING_ENABLED")?.toBoolean() ?: true
```
gabrielittner, line 42:
```suggestion
  var releaseSigningEnabled: Boolean = releaseSigningProperty() ?: true
```

README.md
gabrielittner: In general there are 2 reasons for this
- I think it's generally safer to not add something by default, this way you won't accidentally publish something internal by just running `publish`
- From the plugin perspective we can avoid a lot of edge cases by having less  `afterEvaluate` blocks, we won't be able to get rid of all of them (like when detecting which plugins are applied to the build) but at least some. 

With properties we can still offer easy, low overhead configuration without the cost of afterEvaluate. If you want to publish to maven central you need to add the other properties anyways so it doesn't really increase the setup cost. 

The reason for not using `default` as the value when no property is present is that it won't work well with the DSL. We would then call `publishToMavenCentral(DEFAULT)` immediately and users wouldn't be able to call `publishToMavenCentral(S01)` if the prefer to use the DSL over properties.

vanniktech (approved):

Issue #333, opened 2022-03-02
Javadoc Not Generated in Kotlin Multiplatform Project


cedrickcooke:
New to version `0.19.0`, publishing a Kotlin Multiplatform module that has both JVM and Android targets fails to generate a Javadoc jar for the JVM module.

The gradle publish still succeeds, but Sonatype won't let you close the repository to release it.

gabrielittner:
Which version of AGP are you using?

twyatt:
> Which version of AGP are you using?

7.1.2

eygraber:
Is this caused by https://github.com/vanniktech/gradle-maven-publish-plugin/commit/a91b7655b99ddf3ac8e7f281b0f856fa3b3046bd#diff-f099650db8c1c47f7c5a8a1cbae50bad884723e41109ec858b0f0b9a04e15f8cR60

vanniktech:
I've had the exact same issue here too: https://github.com/vanniktech/Emoji/

AGP: 7.0.4, Kotlin 1.6.21, Dokka 1.6.21 & Maven Publish 0.19.0
Ended up downgrading to Maven Publish 0.18.0 and it seems to be fixed.

This is also easily reproducible when cloning the project, bumping back to Maven Publish 0.19.0 and executing `./gradlew publishToMavenLocal` if you look at the `emoji-ios-jvm` directory there's no javadoc jar file.

gabrielittner:
@eygraber It was caused by that commit but the linked line is intentional, when the Kotlin MPP and Android plugins are applied we only want to run `configure(KotlinMultiplatform(...))` which takes care of configuring publishing for all multiplatform targets.

vanniktech:
Tried this out in https://github.com/vanniktech/Emoji/pull/838/files by doing a `./gradlew publishToMavenLocal` and it does look like everything is nicely generated:

```
➜  vanniktech git:(master) ✗ tree emoji-twitter*
emoji-twitter
├── 0.16.0-SNAPSHOT
│   ├── emoji-twitter-0.16.0-SNAPSHOT-javadoc.jar
│   ├── emoji-twitter-0.16.0-SNAPSHOT-javadoc.jar.asc
│   ├── emoji-twitter-0.16.0-SNAPSHOT-kotlin-tooling-metadata.json
│   ├── emoji-twitter-0.16.0-SNAPSHOT-kotlin-tooling-metadata.json.asc
[...]

PR #350, opened 2022-05-27
fix legacy Android setup in mpp projects

gabrielittner:
This fixes publishing Kotlin Multiplatform projects that apply an Android plugin < 7.1.2. In 0.18.0 and earlier Android publishing was configured in the same when as all other plugins and had lower priority than Kotlin MPP ([see here](https://github.com/vanniktech/gradle-maven-publish-plugin/commit/a91b7655b99ddf3ac8e7f281b0f856fa3b3046bd#diff-f099650db8c1c47f7c5a8a1cbae50bad884723e41109ec858b0f0b9a04e15f8cL29-L32)). The changes for the new Android API moved it out of that when. The code path for the new API had a check to not do anything for multiplatform, but the code path for older AGP versions was missing this which resulted in the Kotlin MPP publication being overridden. This restores the 0.18.0 behavior of not calling the Android APIs in a mpp project.

I will separately look at the issue of MPP + AGP >= 7.1.2 (#333)

### DIFF
--- a/plugin/src/main/kotlin/com/vanniktech/maven/publish/legacy/Platform.kt
+++ b/plugin/src/main/kotlin/com/vanniktech/maven/publish/legacy/Platform.kt
@@ -54,7 +54,7 @@ internal fun Project.configureNotAndroidPlatform() {
 
 internal fun Project.configureAndroidPlatform() {
   if (hasWorkingNewAndroidPublishingApi()) {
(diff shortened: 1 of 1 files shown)


vanniktech (approved):
This is hard to test in an integration test right?

gabrielittner:
I've added a general MPP + Android integration test in #351

PR #351, opened 2022-05-27
fix Kotlin MPP project not being configured when Android plugin is present

### LINKED ISSUE #333: Javadoc Not Generated in Kotlin Multiplatform Project

gabrielittner:
The check for `configureNotAndroidPlatform` would also apply for MPP projects that do apply the Android plugin. Because of that no javadoc jars would be created. As a solution I moved the multiplatform project to a general `plugins.withId`. The general block that we have in `afterEvaluate` will explicitly not handle it anymore and I also moved the check to not handle Android into that again. Maybe we can move the others to `plugins.withId` too but I'm not sure yet if that would cause issues in some combinations.

Closes #333

### DIFF
--- a/plugin/src/integrationTest/kotlin/com/vanniktech/maven/publish/MavenPublishPluginIntegrationTest.kt
+++ b/plugin/src/integrationTest/kotlin/com/vanniktech/maven/publish/MavenPublishPluginIntegrationTest.kt
@@ -258,8 +258,6 @@ class MavenPublishPluginIntegrationTest {
     assertExpectedCommonArtifactsGenerated()
     assertPomContentMatches()
 
-    repoFolder.walkTopDown().forEach { println(it.path) }
-
     val androidArtifactId = "$TEST_POM_ARTIFACT_ID-android"
     assertExpectedCommonArtifactsGenerated(artifactId = androidArtifactId, artifactExtension = "aar")
     // TODO dependency sorting is unstable - the other variants only have 1 and are enough to see that the pom is correct
--- a/plugin/src/integrationTest/kotlin/com/vanniktech/maven/publish/MavenPublishPluginIntegrationTest.kt
+++ b/plugin/src/integrationTest/kotlin/com/vanniktech/maven/publish/MavenPublishPluginIntegrationTest.kt
@@ -177,7 +177,7 @@ class MavenPublishPluginIntegrationTest {
   @Test fun generatesArtifactsAndDocumentationOnKotlinMppProject() {
     setupFixture("passing_kotlin_mpp_project")
 
-    val result = executeGradleCommands(TEST_TASK, "--stacktrace", "--stacktrace")
+    val result = executeGradleCommands(TEST_TASK, "--stacktrace")
 
     assertExpectedTasksRanSuccessfully(result)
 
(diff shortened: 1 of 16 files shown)


plugin/src/main/kotlin/com/vanniktech/maven/publish/legacy/Platform.kt
```diff
@@ -23,21 +23,23 @@ import org.gradle.external.javadoc.StandardJavadocDocletOptions
...
+  }
 }
 
-internal fun Project.configureNotAndroidPlatform() {
+internal fun Project.configureNotAndroidNotMppPlatform() {
   when {
-    plugins.hasPlugin("org.jetbrains.kotlin.multiplatform") ->
-      baseExtension.configure(KotlinMultiplatform(defaultJavaDocOption() ?: JavadocJar.Empty()))
+    plugins.hasPlugin("org.jetbrains.kotlin.multiplatform") -> return // handled separately
+    plugins.hasPlugin("com.android.library") -> return // handled separately
```
vanniktech, line 42:
```suggestion
    plugins.hasPlugin("com.android.library") -> return // Handled separately.
```

plugin/src/main/kotlin/com/vanniktech/maven/publish/legacy/Platform.kt
```diff
@@ -23,21 +23,23 @@ import org.gradle.external.javadoc.StandardJavadocDocletOptions
...
+    configureNotAndroidNotMppPlatform()
+  }
 }
 
-internal fun Project.configureNotAndroidPlatform() {
+internal fun Project.configureNotAndroidNotMppPlatform() {
   when {
-    plugins.hasPlugin("org.jetbrains.kotlin.multiplatform") ->
-      baseExtension.configure(KotlinMultiplatform(defaultJavaDocOption() ?: JavadocJar.Empty()))
+    plugins.hasPlugin("org.jetbrains.kotlin.multiplatform") -> return // handled separately
```
vanniktech, line 41:
```suggestion
    plugins.hasPlugin("org.jetbrains.kotlin.multiplatform") -> return // Handled separately.
```

vanniktech (approved):
Also hard to test this with our integration tests, right?

gabrielittner:
I've added a general integration test that uses Kotlin MPP with Android which should cover at least parts of it.

plugin/src/integrationTest/kotlin/com/vanniktech/maven/publish/MavenPublishPluginIntegrationTest.kt
```diff
@@ -248,6 +248,43 @@ class MavenPublishPluginIntegrationTest {
...
+    setupFixture("passing_kotlin_mpp_with_android_dokka_project")
+
+    val result = executeGradleCommands(TEST_TASK, "--stacktrace", "--stacktrace")
+
+    assertExpectedTasksRanSuccessfully(result, hasDokka = true)
+
+    assertExpectedCommonArtifactsGenerated()
+    assertPomContentMatches()
+
+    repoFolder.walkTopDown().forEach { println(it.path) }
```
vanniktech, line 261: Debugging purposes or should we keep it?

vanniktech (approved):
Thanks a lot @gabrielittner

plugin/src/integrationTest/kotlin/com/vanniktech/maven/publish/MavenPublishPluginIntegrationTest.kt
gabrielittner, line 261:
```suggestion
```

gabrielittner: That was for debugging

PR #354, opened 2022-06-02
remove support for publishing with the old Android Gradle APIs

gabrielittner:
- minimum supported AGP version is now 7.1.2
- also bumped the minimum supported Gradle version to 7.2.0 which is AGP's minimum Gradle version
- removed publishing setup for the old AGP publishing
- removed the deprecated `androidVariantToPublish` from the legacy extension
- removed the deprecated `AndroidLibrary` platform for the base plugin
- cleaned up the tests to use the `7_1_1` versions by default

After also doing the cleanup for the rest of the old extension (the things we deprecated in 0.20.0) I want to look rewrite our tests to be more parameterized so that we don't need as many fixtures and can test across multiple AGP/Kotlin/Gradle versions.


vanniktech (approved):
Can we throw an error our own error when you're using this new stuff with an AGP version less then 7.1.2? 

Just in case someone updates with an old version of AGP

gabrielittner:
We throw an error when applying the plugin with incompatible agp versions https://github.com/vanniktech/gradle-maven-publish-plugin/pull/354/files#diff-95cb06cb11df2913ff9e544f282b3022cc7bf83de866219a8f08021d89a38415R16

PR #384, opened 2022-08-17
Disable sources tasks for KMP project if sources are disabled

carlonzo:
Source artifact is added automatically by the KMP plugin. Manually removing the source artifact breaks Gradle Metadata which throw an exception. 

A workaround would be to disable the task generating the sources, or replace it with a task generating an empty source.
I decided to just disable the task given that Gradle doesnt allow me to replace an already added task. 

I am aware this is a workaround.


vanniktech:
Is this something we want @gabrielittner ?

gabrielittner:
I would prefer if this would happen in the Kotlin plugin itself. They are the ones creating the source jars and adding them to the publications. Please open an issue on youtrack

PR #398, opened 2022-09-03
use properties in closeAndReleaseRepository and move it

gabrielittner:
The task now lives in the plugin module which makes `nexus` free of Gradle. The input for the task is now done through Gradle properties which simplifies it a bit and also means that we don't read the username/password properties during configuration time.


vanniktech (approved):
Nice

PR #399, opened 2022-09-03
replace root plugin with a BuildService that stores whether repo was closed

gabrielittner:
Until now we had a root plugin through which we would create the closeAndReleaseRepository task to ensure that there is just one even in multi project builds. This completely removes that. Instead each project gets its own closeAndReleaseRepository instance and we register a `BuildService` when maven central is configured. This build service is passed to the task and the task will set a variable on it to true after the closing happened. The task will also check that boolean before doing anything and will just return early if that was the case. This way no matter how many close and release tasks there only one of them will actually do something. With the combination of delcationg `usesService` for the task and setting `maxParallelUsages` to 1 Gradle will guarantee that none of the close tasks will run in parallel.

The build service will also be used by the task to create new repositories so I moved all the nexus config that was in the close task over to it.


plugin/src/main/kotlin/com/vanniktech/maven/publish/MavenPublishBaseExtension.kt
```diff
@@ -11,8 +13,8 @@ import org.gradle.api.publish.maven.MavenPublication
 import org.gradle.plugins.signing.SigningPlugin
 
 @Incubating
-abstract class MavenPublishBaseExtension(
-  private val project: Project
+abstract class MavenPublishBaseExtension internal constructor(
```
gabrielittner, line 16:
```suggestion
abstract class MavenPublishBaseExtension(
```

vanniktech (approved):
Interesting, did not know about this API!

PR #401, opened 2022-09-04
add task to create staging repository

gabrielittner:
The new `createStagingRepository` task will make the neccessary calls to create a staging repo on Sonatype and then stores the id of that in both the build service and and a property that is used to compute the url to publish to. When the build service already contains an id the task will just set that on the property so that we only create a single repository for the whole build. In the case of snapshots the task will just return immediately. The Gradle publish tasks depend on the new task so that it's automatically running before we upload anything. 

With this just running `./gradlew publish` or `./gradlew publish*ToMavenCentralRepository`will automatically create a staging repo and upload all artifacts to that one repo.

One note is that `closeAndReleaseRepository` is not integrated with this yet. So you need a second Gradle invocation (`./gradlew closeAndReleaseRepository`) to do that.

### DIFF
--- a/plugin/src/main/kotlin/com/vanniktech/maven/publish/MavenPublishBaseExtension.kt
+++ b/plugin/src/main/kotlin/com/vanniktech/maven/publish/MavenPublishBaseExtension.kt
@@ -1,6 +1,7 @@
 package com.vanniktech.maven.publish
 
 import com.vanniktech.maven.publish.sonatype.CloseAndReleaseSonatypeRepositoryTask.Companion.registerCloseAndReleaseRepository
+import com.vanniktech.maven.publish.sonatype.CreateSonatypeRepositoryTask.Companion.registerCreateRepository
 import com.vanniktech.maven.publish.sonatype.SonatypeRepositoryBuildService.Companion.registerSonatypeRepositoryBuildService
 import java.util.concurrent.Callable
 import org.gradle.api.Action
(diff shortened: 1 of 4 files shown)


plugin/src/main/kotlin/com/vanniktech/maven/publish/MavenPublishBaseExtension.kt
```diff
@@ -52,12 +48,25 @@ abstract class MavenPublishBaseExtension(
...
+    val stagingRepositoryId = createRepository.flatMap { it.stagingRepositoryId }
+
     project.gradlePublishing.repositories.maven { repo ->
       repo.name = "mavenCentral"
       repo.setUrl(sonatypeHost.map { it.publishingUrl(project.versionIsSnapshot, stagingRepositoryId) })
       repo.credentials(PasswordCredentials::class.java)
     }
 
+    project.gradlePublishing.publications.withType(MavenPublication::class.java).all { publication ->
+      project.afterEvaluate {
```
vanniktech, line 63: Why do we need the afterEvaluate here?

plugin/src/main/kotlin/com/vanniktech/maven/publish/sonatype/CreateSonatypeRepositoryTask.kt
```diff
@@ -0,0 +1,64 @@
...
+
+  @TaskAction
+  fun createStagingRepository() {
+    if (versionIsSnapshot.get()) {
+      return
+    }
+
+    val service = this.buildService.get()
+
+    // if repository was already create in this build this is a no-op
```
vanniktech, line 33:
```suggestion
    // if repository was already created in this build this is a no-op
```

vanniktech (approved):

plugin/src/main/kotlin/com/vanniktech/maven/publish/MavenPublishBaseExtension.kt
gabrielittner: The task is only created at some point after we get the callback here when a new publication was added. Afaik there is no version of `tasks.named(...)` that will give you a callback once a task with that name is created. So it's either `afterEvaluate` or doing `tasks.configureEach` and then filtering by name and in that case I think afterEvaluate is preferable.

PR #403, opened 2022-09-04
automatically close the created repository and optionally also release it

gabrielittner:
This will automatically close the repository that we created at the end of the build and based on a Gradle property or a parameter in the DSL also release it automatically. I think we don't need an option to control the automatic closing, each build will create a new repository anyways so there shouldn't be anything added after the fact. If there is a valid use case for it we can still make it optional later. Releasing on the other hand is a bigger step so I decided to make it opt-in. Even if you want to use it it allows for something like only enabling it on CI by setting an env var.

The closing and releasing is not done through the `closeAndReleaseRepository` task. The issue with the task is that it would need to wait for publishing in all other subprojects to finish so we would need cross project dependencies. One solution would have been the root plugin that we had before which would at least simplify it to only the root project task having to wait for all projects to finish publishing, but I also had some other issues that plugin. The solution is to trigger this from the build service. When a build service implements `AutoCloseable` Gradle will call `close` when the build service is no longer needed. Implementing `OperationCompletionListener` and registering it will make it so that we can detect that no task failed (in that case close and release is not run) and will make it so that the build service is needed for the whole build which means that `close` is only called after the last task finished. The `close` implementation then checks if a staging repo was created and then does the close and release operations.

I'm not 100% sure yet about the future of the close and release task. For now I made a small change to the nexus code so that running it on an already close repo will simply release it instead of printing an error that it is not open. So it can be used to then release it manually if automatic releasing isn't enabled. It could also be used to manually retry closing and releasing the already created staging repo if it failed. So it probably continue existing. One future change could be making the `--release` cli parameter required though and removing automatically finding a repo. The build will print the id of the repo when creating and automatically closing it, so you always know it.

### DIFF
--- a/plugin/src/main/kotlin/com/vanniktech/maven/publish/sonatype/SonatypeRepositoryBuildService.kt
+++ b/plugin/src/main/kotlin/com/vanniktech/maven/publish/sonatype/SonatypeRepositoryBuildService.kt
@@ -7,12 +7,16 @@ import org.gradle.api.provider.Provider
 import org.gradle.api.services.BuildService
 import org.gradle.api.services.BuildServiceParameters
 import org.gradle.api.services.BuildServiceRegistry
+import org.gradle.tooling.events.FinishEvent
+import org.gradle.tooling.events.OperationCompletionListener
+import org.gradle.tooling.events.task.TaskFailureResult
 
-internal abstract class SonatypeRepositoryBuildService : BuildService<SonatypeRepositoryBuildService.Params> {
+internal abstract class SonatypeRepositoryBuildService : BuildService<SonatypeRepositoryBuildService.Params>, AutoCloseable, OperationCompletionListener {
   internal interface Params : BuildServiceParameters {
     val sonatypeHost: Property<SonatypeHost>
     val repositoryUsername: Property<String>
     val repositoryPassword: Property<String>
+    val automaticRelease: Property<Boolean>
   }
 
   val nexus = Nexus(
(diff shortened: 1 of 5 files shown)


plugin/src/main/kotlin/com/vanniktech/maven/publish/sonatype/SonatypeRepositoryBuildService.kt
```diff
@@ -36,19 +40,43 @@ internal abstract class SonatypeRepositoryBuildService : BuildService<SonatypeRe
...
+  override fun close() {
+    if (buildHasFailure) {
+      return
+    }
+
+    val stagingRepositoryId = this.stagingRepositoryId
+    if (stagingRepositoryId != null) {
+      nexus.closeStagingRepository(stagingRepositoryId)
+      if (parameters.automaticRelease.get()) {
+        nexus.releaseStagingRepository(stagingRepositoryId)
```
vanniktech, line 60: Isn't this a change from what we had before when executing `closeAndReleaseRepository`. With 0.21.0 it always does close and release. And now when upgrading to this version it only does `close`. Why don't we auto release by default?

gabrielittner: This is how it compares:

Running `publish`:
0.21.0: implictly createst staging repo, uploads
now - flag off (default): creates staging repo, uploads, closes it
now - flag on: creates staging repo, uploads, closes it, releases it

Running `closeAndReleaseRepository` when repo is open
0.21.0: finds repo, if there is just one closes and releases it
now: finds repo, if there is just one closes and releases it

Running `closeAndReleaseRepository` when repo is closed
0.21.0: finds repo, fails
now: finds repo, if there is just one releases it


So `closeAndReleaseRepository` generally behaves like before and if you had `publish` and `closeAndReleaseRepository` running on ci (or you were manually executing them) the end result is the same. The automatic release flag gives you the option to let publish also release it so that you don't need to run `closeAndReleaseRepository` at all. I thought making the automatic release opt in is a bit safer as it's not something you can reverse.

vanniktech: Aaah okay. Now i understand it, thanks!

vanniktech (approved):

PR #404, opened 2022-09-04
workaround issues with the signing setup

### LINKED ISSUE #338: Gradle detecting problems with publish and sign dependencies for a multiplatform project
If I run a publish task for a module which depends on another module
```
Gradle detected a problem with the following location: '~/Development/sqldelight/runtime/build/libs/runtime-2.1.0-SNAPSHOT-javadoc.jar.asc'. Reason: Task ':runtime:publishIosArm32PublicationToInstallLocallyRepository' uses this output of task ':runtime:signIosX64Publication'
```

This _might_ actually be a bug with gradle or kotlin but its hard to tell. It looks like the publish task references the jar signature file, which got signed by a previous task, so gradle expects there to be a task dependency?

### LINKED ISSUE #387: Incompatible with 1.0.0 of Gradle's `com.gradle.plugin-publish`plugin?
Use of `com.gradle.plugin-publish` is required to publish plugins to Gradle's plugin repository.  As of the recently-released version  1.0.0, it seems that `gradle-maven-publish-plugin` is incompatible with Gradle's own plugin:

```
* Exception is:
org.gradle.api.ProjectConfigurationException: A problem occurred configuring root project 'dexcount-gradle-plugin'.
	at org.gradle.configuration.project.LifecycleProjectEvaluator.wrapException(LifecycleProjectEvaluator.java:84)
	at org.gradle.configuration.project.LifecycleProjectEvaluator.addConfigurationFailure(LifecycleProjectEvaluator.java:77)
	...
Caused by: org.gradle.api.internal.tasks.DefaultTaskContainer$DuplicateTaskException: Cannot add task 'signDexcountPluginMarkerMavenPublication' as a task with that name already exists.
	at org.gradle.api.internal.tasks.DefaultTaskContainer.failOnDuplicateTask(DefaultTaskContainer.java:257)
	...
	at org.gradle.plugins.signing.SigningExtension.sign(SigningExtension.java:395)
	at com.vanniktech.maven.publish.MavenPublishBaseExtension.signAllPublications(MavenPublishBaseExtension.kt:114)
	at com.vanniktech.maven.publish.MavenPublishPlugin.apply$lambda-0(MavenPublishPlugin.kt:65)
        ...
```

Interestingly, swapping the order of plugins in the `plugins` block so that this plugin comes before Gradle's, works around the issue.  Would it be possible for this plugin to adapt to the presence of the Gradle plugin?  Or should I be filing an issue with them instead?

gabrielittner:
closes #387
closes #338

### DIFF
--- a/plugin/src/main/kotlin/com/vanniktech/maven/publish/MavenPublishBaseExtension.kt
+++ b/plugin/src/main/kotlin/com/vanniktech/maven/publish/MavenPublishBaseExtension.kt
@@ -11,8 +11,11 @@ import org.gradle.api.credentials.PasswordCredentials
 import org.gradle.api.provider.Property
 import org.gradle.api.publish.maven.MavenPom
 import org.gradle.api.publish.maven.MavenPublication
+import org.gradle.api.publish.maven.tasks.AbstractPublishToMaven
+import org.gradle.api.publish.maven.tasks.PublishToMavenRepository
 import org.gradle.build.event.BuildEventsListenerRegistry
 import org.gradle.configurationcache.extensions.serviceOf
+import org.gradle.plugins.signing.Sign
 import org.gradle.plugins.signing.SigningPlugin
 
 @Incubating
@@ -69,11 +72,9 @@ abstract class MavenPublishBaseExtension(
       repo.credentials(PasswordCredentials::class.java)
     }
 
-    project.gradlePublishing.publications.withType(MavenPublication::class.java).all { publication ->
-      project.afterEvaluate {
-        project.tasks.named("publish${publication.name.capitalize()}PublicationToMavenCentralRepository") {
-          it.dependsOn(createRepository)
-        }
+    project.tasks.withType(PublishToMavenRepository::class.java).configureEach { publishTask ->
+      if (publishTask.name.endsWith("ToMavenCentralRepository")) {
+        publishTask.dependsOn(createRepository)
       }
     }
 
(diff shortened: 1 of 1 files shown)


plugin/src/main/kotlin/com/vanniktech/maven/publish/MavenPublishBaseExtension.kt
```diff
@@ -69,11 +71,9 @@ abstract class MavenPublishBaseExtension(
...
     }
 
-    project.gradlePublishing.publications.withType(MavenPublication::class.java).all { publication ->
-      project.afterEvaluate {
-        project.tasks.named("publish${publication.name.capitalize()}PublicationToMavenCentralRepository") {
-          it.dependsOn(createRepository)
-        }
+    project.tasks.withType(PublishToMavenRepository::class.java).configureEach { publishTask ->
+      if (publishTask.name.endsWith("ToMavenCentralRepository")) {
+        publishTask.dependsOn(createRepository)
```
gabrielittner, line 76: Found out that `PublishToMavenRepository` is in the public API so we can use it here as a filter which makes it nicer than afterEvaluate

plugin/src/main/kotlin/com/vanniktech/maven/publish/MavenPublishBaseExtension.kt
```diff
@@ -114,14 +114,27 @@ abstract class MavenPublishBaseExtension(
...
     }
+
+    // TODO: replace with the following line after https://github.com/gradle/gradle/issues/21857 is fixed
+    //  project.gradleSigning.sign(project.gradlePublishing.publications)
+    project.gradlePublishing.publications.withType(MavenPublication::class.java).all { publication ->
+      val task = project.tasks.findByName("sign${publication.name.capitalize()}Publication")
+      if (task == null) {
+        project.gradleSigning.sign(publication)
+      }
+    }
```
gabrielittner, line 132: Only call sign on a publication if there is no signing task for it yet which fixes the issue when other plugins already called it.

plugin/src/main/kotlin/com/vanniktech/maven/publish/MavenPublishBaseExtension.kt
```diff
@@ -114,14 +115,27 @@ abstract class MavenPublishBaseExtension(
...
+
+    // TODO: remove after https://youtrack.jetbrains.com/issue/KT-46466 is fixed
+    project.tasks.withType(AbstractPublishToMaven::class.java) { publishTask ->
+      publishTask.dependsOn(project.tasks.withType(Sign::class.java))
+    }
```
gabrielittner, line 138: Workaround taken from the linked ticket

vanniktech (approved):

Issue #381, opened 2022-08-09
`closeAndReleaseRepository` fails because `publish` creates multiple staging repositories


eygraber:
I have a CI task that runs `./gradlew publish && ./gradlew closeAndReleaseRepository`.

This usually works, but I just ran into [a CI error](https://github.com/eygraber/gradle-utils/runs/7738714637?check_suite_focus=true#step:7:265):

```
More than 1 matching staging repository found in account ***. You can can explicitly choose one by passing it as an option like this "./gradlew closeAndReleaseRepository --repository comexample-123". Available repositories are: com***-1119, com***-1120
```

I confirmed that there were two staging repositories created, and they were both created by the `publish` task.

vanniktech:
I also had this once but it was since I was doing 2 releases of different repositories simultaneously.

eygraber:
In this case there was only 1 release happening.

It looks like the [gradle-nexus](https://github.com/gradle-nexus/publish-plugin/blob/master/src/main/kotlin/io/github/gradlenexus/publishplugin/InitializeNexusStagingRepository.kt) creates a staging repo for each build, but you have to use their lifecycle task instead of the built in gradle ones.

This plugin could accomplish something similar, while retaining the ability to just use the gradle publish tasks by:

1. Make an initialization task that creates a staging repo
  a. Call `GET profiles` and find the `name` that matches the project's `group` and get its `id`
  b. Call `POST profiles/<id>/start` to create a repo, using the project's GAV or a unique user override for the `description` field
  c. Get the `stagedRepositoryId` from the response, and use that to set the repo url (via a property)
2. Make the gradle publish tasks depend on the initialization task
3. Make the `closeAndReleaseRepository` task depend on the gradle publish tasks 
  a. Call `GET profile_repositories` and find the matching description
  b. If there is 0 or more than 1 matching, error
  c. Otherwise close and release :smile:

eygraber:
In any case, the specific issue here seems to be that after adding a new gradle project, all artifacts are put into one staging repo, and one pom file from one of the projects are put into a separate staging repo.

I have no idea what is causing this. I reached out to Sonatype to see if it is an issue on their end. The only clue I have is that the pom file is always from a gradle plugin that I'm publishing.

eygraber:
And now I ran it a bunch of time and it suddenly worked ¯\_(ツ)_/¯

gabrielittner:
[...]

PR #405, opened 2022-09-05
Create nexus lazily to avoid an error when properties aren't set

eygraber:
Fixes https://github.com/vanniktech/gradle-maven-publish-plugin/issues/381#issuecomment-1236436712

The other way to do it would be to make `Nexus` (and all subsequent types) take `Provider<String>` instead of `String`


vanniktech (approved):

gabrielittner (approved):
I wonder why I didn't run into the issue while testing. Thanks for trying out the snapshots and for submitting the fix

PR #408, opened 2022-09-08
Use composite builds instead of buildSrc

chachako:
Inspired by [Stop using Gradle buildSrc. Use composite builds instead](https://proandroiddev.com/stop-using-gradle-buildsrc-use-composite-builds-instead-3c38ac7a2ab3) :)

After:
image


gabrielittner (approved):
Thanks for the contribution

PR #410, opened 2022-09-08
drop repository when build fails

### LINKED ISSUE #406: Drop an automatically created sonatype staging repo if there is a build failure
If a sonatype staging repo is automatically created, but the build ends up failing for some reason, there will be a buildup of empty staging repos in nexus. If those could be dropped when the build fails that would be very helpful.

gabrielittner:
Makes an attempt to drop the created staging repository in case of build failures. This is on a best effort basis and a failure to drop the repo will be silently ignored.

closes #406

### DIFF
--- a/plugin/src/main/kotlin/com/vanniktech/maven/publish/sonatype/SonatypeRepositoryBuildService.kt
+++ b/plugin/src/main/kotlin/com/vanniktech/maven/publish/sonatype/SonatypeRepositoryBuildService.kt
@@ -2,14 +2,15 @@ package com.vanniktech.maven.publish.sonatype
 
 import com.vanniktech.maven.publish.SonatypeHost
 import com.vanniktech.maven.publish.nexus.Nexus
+import java.io.IOException
 import org.gradle.api.provider.Property
 import org.gradle.api.provider.Provider
 import org.gradle.api.services.BuildService
 import org.gradle.api.services.BuildServiceParameters
 import org.gradle.api.services.BuildServiceRegistry
+import org.gradle.tooling.events.FailureResult
 import org.gradle.tooling.events.FinishEvent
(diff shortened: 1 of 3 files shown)


plugin/src/main/kotlin/com/vanniktech/maven/publish/sonatype/SonatypeRepositoryBuildService.kt
```diff
@@ -42,24 +43,26 @@ internal abstract class SonatypeRepositoryBuildService : BuildService<SonatypeRe
...
-        nexus.releaseStagingRepository(stagingRepositoryId)
+      if (buildIsSuccess) {
+        nexus.closeStagingRepository(stagingRepositoryId)
+        if (parameters.automaticRelease.get()) {
+          nexus.releaseStagingRepository(stagingRepositoryId)
+        }
+      } else {
+        try {
+          nexus.dropStagingRepository(stagingRepositoryId)
+        } catch (_: IOException) {}
```
vanniktech, line 65: Should we log it?

vanniktech (approved):

plugin/src/main/kotlin/com/vanniktech/maven/publish/sonatype/SonatypeRepositoryBuildService.kt
gabrielittner: When the build fails people often jump to the last message that looks like an error which in this case would be misleading because this didn't fail the build. Maybe as an info level log? Then you could find out about it when running with `--info` but by default Gradle doesn't print it

vanniktech: Sounds good!

PR #423, opened 2022-09-20
Avoid configuring tasks unnecessarily

3flex:


plugin/src/main/kotlin/com/vanniktech/maven/publish/tasks/JavadocJar.kt
```diff
@@ -31,12 +32,21 @@ open class JavadocJar : Jar() {
       }
     }
 
-    private fun Project.dokkaJavadocJar(options: JavadocJarOption.Dokka): TaskProvider<*> {
+    private fun Project.dokkaJavadocJar(): TaskProvider<*> {
       return tasks.register("dokkaJavadocJar", JavadocJar::class.java) {
-        val task = tasks.named(options.taskName)
+        val task = provider { findDokkaTask() }
```
3flex, line 37: This is the key enabling change in the second commit - the rest is refactoring.

vanniktech (approved):

Goooler (approved):

plugin/src/main/kotlin/com/vanniktech/maven/publish/Platform.kt
```diff
@@ -337,7 +336,7 @@ sealed class JavadocJar {
    * Creates a javadoc jar using Dokka's output. The argument is the name of the dokka task that should be used
    * for that purpose.
    */
-  data class Dokka(val taskName: String) : JavadocJar()
+  object Dokka : JavadocJar()
```
gabrielittner, line 339: This is also used as a public API for the base plugin to let users sepecify manually which dokka task should be used as input for the javadoc jar. I'd suggest at least having the following to still have that as a possibility (the base plugin as a principle should never do/decide anything automatically) and keep the automatic finding logic in the main plugin

```
data class Dokka(val taskName: Provider<String>) : JavadocJar()
```

This would still be a breaking change though

gabrielittner (changes requested):

plugin/src/main/kotlin/com/vanniktech/maven/publish/Platform.kt
gabrielittner: Maybe, this to make it not breaking
```
data class Dokka private constructor(
  internal val taskName: Any
) : JavadocJar() {
  constructor(val taskName: String) : this(taskName)
  constructor(val taskName: Provider<String>) : this(taskName)
}
```

vanniktech (approved):

PR #427, opened 2022-09-28
Update README.me with publish task behaviour changes

Orgmir:
Running `./gradlew tasks` after setting up `com.vanniktech.maven.publish` did not show the Maven Central publishing tasks. Since version 0.21.0 publishing to Maven Central is disabled until we add:
```properties
SONATYPE_HOST=DEFAULT
RELEASE_SIGNING_ENABLED=true
```

I have updated the README.md with the info from the CHANGELOG.md, hopefully it will help people setup their repositories correctly without wasting time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,9 +66,37 @@ the rest by putting them into `<project-dir>/gradle.properties`.
 
 ### Where to upload to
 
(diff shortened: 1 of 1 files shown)


gabrielittner:
Thank you for the contribution, I've completely rewritten the docs and also made this part clearer

PR #442, opened 2022-10-21
start rewriting integration tests

gabrielittner:
This should make integration tests more scalable by not requiring to set up a full fixture for each scenario. Instead a project is defined with `ProjectSpec` and `ProjectSpecRunner` will then dynamically create a matching project. Together with `TestOptions` this allows us to also cover more cases than before:
- test both the dsl and the properties config
- test without signing, gpg file based signging and in memory signing
- test with different Gradle, AGP and Kotlin versions

Since this produces a lot of different combinations I limited the version to minimum supported version, current version, and preview versions (latest patch version of each).

The old tests still exist for now because I still need to implement checking the pom content in these new tests. I've added a comment to them to easily identify whether they were already migrated or still need to be.


plugin/src/integrationTest/kotlin/com/vanniktech/maven/publish/ProjectSpecs.kt
```diff
@@ -0,0 +1,119 @@
...
+        namespace "com.test.library"
+        compileSdkVersion 29
+    }
+
+    // disable the dokka task to speed up Android tests significantly
+    afterEvaluate {
+      tasks.named("javaDocReleaseGeneration").configure {
+        it.enabled = false
+      }
+    }
```
gabrielittner, line 107: This isn't great but it will still produce an empty javadoc jar and our tests are only checking for existence of them and never look at the content.

Disabled task:
Screenshot 2022-10-21 at 14 55 35
Enabled task:
Screenshot 2022-10-21 at 15 37 40

plugin/src/integrationTest/kotlin/com/vanniktech/maven/publish/ProjectSpec.kt
```diff
@@ -0,0 +1,27 @@
...
+import org.gradle.testkit.runner.BuildResult
+
+data class ProjectSpec(
+  val plugins: List<PluginSpec>,
+  val group: String,
+  val artifactId: String,
+  val version: String,
+  val properties: Map<String, String>,
+  val sourceFiles: List<Pair<String, String>>,
+  val buildFileExtra: String = ""
```
vanniktech, line 13:
```suggestion
  val buildFileExtra: String = "",
```

plugin/src/integrationTest/kotlin/com/vanniktech/maven/publish/TestOptions.kt
```diff
@@ -0,0 +1,68 @@
...
+  val value: String,
+  val minGradleVersion: GradleVersion,
+  val firstUnsupportedGradleVersion: GradleVersion? = null,
+) {
+  AGP_7_1(
+    value = "7.1.2",
+    minGradleVersion = GradleVersion.GRADLE_7_2,
+    firstUnsupportedGradleVersion = GradleVersion.GRADLE_8_0,
+  ),
+  // AGP_7_2(
```
vanniktech, line 29: Let's remove this if we don't use it?

plugin/src/integrationTest/kotlin/com/vanniktech/maven/publish/TestOptions.kt
```diff
@@ -0,0 +1,68 @@
...
+    minGradleVersion = GradleVersion.GRADLE_7_4,
+  ),
+  AGP_8_0(
+    value = "8.0.0-alpha05",
+    minGradleVersion = GradleVersion.GRADLE_7_5,
+  ),
+}
+
+enum class KotlinVersion(val value: String) {
+  // KT_1_5("1.5.32"),
```
vanniktech, line 49: Same?

plugin/src/integrationTest/kotlin/com/vanniktech/maven/publish/TestOptions.kt
```diff
@@ -0,0 +1,68 @@
...
+  GRADLE_7_2("7.2"),
+  // GRADLE_7_3("7.3.3"),
+  // GRADLE_7_4("7.4.2"),
+  GRADLE_7_5("7.5.1"),
+  GRADLE_7_6("7.6-milestone-1"),
+  GRADLE_8_0("8.0-milestone-1"),
+  ;
+
+  companion object {
+    // aliases for the skipped version to be able to reference the correct one in AgpVersion
```
vanniktech, line 64: Skipped because?

vanniktech (approved):
🚀

vanniktech (approved):

PR #461, opened 2022-12-20
add pom validation to new integration tests

gabrielittner:
I'm using the Apache Maven model library for this. Unfortunately the models don't implement equals so we can't just check that the models are equal and still need to resort to string representations of the written XML. The other issue is that even though Gradle is using the same library under the hood, Gradle's output is inconsitent between project types, for example I had this diff in one of the tests but not in another

```diff
-<project xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns="http://maven.apache.org/POM/4.0.0"
-    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+<project xmlns="http://maven.apache.org/POM/4.0.0" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
```
Both are the same 3 elements but differently sorted and one with a line break.

So I ended up also parsing the pom created by Gradle and then turning it into XML again. The only thing that is getting lost through this is the comment that Gradle adds to the XML that references the module metadata, but I don't think that that is too important for these tests.

I also removed the already migrated old tests since the only missing part was checking the pom.


plugin/src/integrationTest/kotlin/com/vanniktech/maven/publish/Subjects.kt
```diff
@@ -176,3 +182,36 @@ class SourcesJarSubject private constructor(
...
+  failureMetadata: FailureMetadata,
+  private val artifact: Path,
+  private val result: ProjectResult,
+) : ArtifactSubject(failureMetadata, artifact) {
+
+  companion object {
+    private val BUILD_RESULT_SUBJECT_FACTORY: Factory<PomSubject, Pair<Path, ProjectResult>> =
+      Factory { metadata, actual -> PomSubject(metadata, actual.first, actual.second) }
+
+    @JvmStatic
```
vanniktech, line 196: Do we need this? I thought we are Kotlin only

vanniktech (approved):
Great work

plugin/src/integrationTest/kotlin/com/vanniktech/maven/publish/Subjects.kt
gabrielittner: We don't just came from copying the default java sample

gabrielittner: [67a10ea](https://github.com/vanniktech/gradle-maven-publish-plugin/pull/461/commits/67a10ea8b42bf19c191931743e8ddbe48ef5bf4c)

PR #466, opened 2022-12-20
add helper to access mavenPublications

gabrielittner:


plugin/src/main/kotlin/com/vanniktech/maven/publish/MavenPublishBaseExtension.kt
```diff
@@ -123,12 +123,12 @@ abstract class MavenPublishBaseExtension(
...
-    project.gradlePublishing.publications.withType(MavenPublication::class.java).all { publication ->
-      val task = project.tasks.findByName("sign${publication.name.capitalize()}Publication")
-      if (task == null) {
-        project.gradleSigning.sign(publication)
+    if (GradleVersion.current() < GradleVersion.version("8.0")) {
+      project.mavenPublications { publication ->
+        val task = project.tasks.findByName("sign${publication.name.capitalize()}Publication")
+        if (task == null) {
+          project.gradleSigning.sign(publication)
+        }
```
gabrielittner, line 131: Issue is fixed in 8.0 so we don't need to workaround that causes eager task creation anymore

plugin/src/main/kotlin/com/vanniktech/maven/publish/MavenPublishBasePlugin.kt
```diff
@@ -3,14 +3,13 @@ package com.vanniktech.maven.publish
 import org.gradle.api.Plugin
 import org.gradle.api.Project
 import org.gradle.api.publish.maven.plugins.MavenPublishPlugin as GradleMavenPublishPlugin
-import org.gradle.util.VersionNumber
+import org.gradle.util.GradleVersion
```
gabrielittner, line 6: `VersionNumber` will be removed in Gradle 9

plugin/src/main/kotlin/com/vanniktech/maven/publish/Platform.kt
```diff
@@ -76,11 +76,9 @@ data class GradlePlugin @JvmOverloads constructor(
 ) : Platform() {
 
   override fun configure(project: Project) {
-    project.gradlePublishing.publications.withType(MavenPublication::class.java).all {
-      if (it.name == "pluginMaven") {
-        it.withSourcesJar { project.javaSourcesJar(sourcesJar) }
-        it.withJavadocJar { project.javadocJarTask(javadocJar) }
-      }
+    project.mavenPublicationsWithoutPluginMarker {
```
gabrielittner, line 79: The check is the same, just the other way around

vanniktech (approved):

PR #467, opened 2022-12-21
migrate multiplatform tests

gabrielittner:


gabrielittner:
These take way too long to run but I'll take a look at it separately.

PR #473, opened 2022-12-23
add the ability to set coordinates to the base plugin

### LINKED ISSUE #326: Base plugin does not allow configuring artifact ID
I was just trying to migrate my project to use this plugin.  But unfortunately, I couldn't find a way to configure the artifactId without going through hoops, and the default project name doesn't really work for my use case.
Ideally, I'd like to be able to just configure an `artifactId` field on the `MavenPublishBaseExtension` directly, or perhaps, via the Platform API (like` AndroidLibrary(..., artifactId = xxx)` for example).

I wonder if there was any technical reason behind this, or if it was just an oversight in the library

### LINKED ISSUE #445: `createStagingRepository` fails when group is only set via `MavenPublication.groupId`
This is new behavior in `0.22.0`, using the base plugin.  The Gradle version is `7.5.1`, but that shouldn't matter.

I don't have a full stacktrace at the moment but the exception message made it pretty clear what's happening:

> No matching staging profile found in account rbusarow. It is expected that the account contains a staging profile that matches or is the start of workflow.workflow-config.Available profiles are: com.squareup

The `createStagingRepository` task is using `project.group` as the groupId.

I have been setting `groupId` alongside `artifactId` like this:

```kotlin
configure<PublishingExtension> {
  publications.withType(MavenPublication::class.java) pub@{
    this@pub.groupId = "com.example"
    this@pub.artifactId = "some-id"
  }
}
```

This is done in order to an old conflict resolution issue in Gradle (https://github.com/gradle/gradle/issues/847).  To summarize, a unique `project.group` is necessary to distinguish project dependencies when they have the same simple name, like `:a:api` and `:b:api`.

But now with 0.22.0, `MavenPublishBaseExtension` is hard-coded to get the group from the project:

https://github.com/vanniktech/gradle-maven-publish-plugin/blob/2e12d99d80fac3d7b4a0f839757d8fb1ffb5dc5a/plugin/src/main/kotlin/com/vanniktech/maven/publish/MavenPublishBaseExtension.kt#L64-L66

Ideally, the extension and `CreateSonatypeRepositoryTask` would also use the publication's `groupId` value.

gabrielittner:
closes #326 and closes #445 

This moves the last bit that was not in the base plugin/extension to it. The main reason why I didn't do this before was because the artifact id is a bit hacky but in the end it is fine. There are 2 behavior changes here which are required to fully resolve the issue described in #445:
- We don't set `project.group` and `project.version` anymore. The main reason the plugin did this is because those 2 are the default group and version so we set those and didn't need to do anything else. Since we are now setting them explicitly anyways there is no need for this anymore. Afaik this will also not have any effects on users of the plugin
- It is not possible anymore to override the used group/version by setting `project.group` and `project.version` if also the matching Gradle property is present (this change is the removed test case). If someone relied on this change will affect them however I don't really see this as a common thing or use case. 

`project.group` and `project.version` will continue to be the default if neither the properties exist or the new `coordinates` API was called. So anyone following the DSL guide or old base plugin README is not required to make any changes, but they might still want to use the new API instead.


plugin/src/integrationTest/kotlin/com/vanniktech/maven/publish/ProjectSpecRunner.kt
```diff
@@ -124,12 +125,7 @@ private fun ProjectSpec.publishingBlock(options: TestOptions): String {
   }
 }
 
-private fun ProjectSpec.writeSettingFile(path: Path, options: TestOptions) {
-  val rootProjectName = if (options.config == TestOptions.Config.DSL) {
-    artifactId
-  } else {
-    "default-root-project-name"
-  }
```
gabrielittner, line 132: This was a workaround to not beeing able to set the artifact id through the dsl which is now not needed anymore.

plugin/src/main/kotlin/com/vanniktech/maven/publish/MavenPublishBaseExtension.kt
```diff
@@ -18,13 +18,16 @@ import org.gradle.plugins.signing.Sign
 import org.gradle.plugins.signing.SigningPlugin
 import org.gradle.util.GradleVersion
 
-@Incubating
 abstract class MavenPublishBaseExtension(
```
gabrielittner, line 21: No need to have the extension itself still has incubating, some of it's methods already don't have it anymore.

plugin/src/main/kotlin/com/vanniktech/maven/publish/MavenPublishBaseExtension.kt
```diff
@@ -138,6 +140,68 @@ abstract class MavenPublishBaseExtension(
...
+    project.mavenPublicationsWithoutPluginMarker {
+      // the multiplatform plugin creates its own publications, so it is ok to use hasPlugin in here
+      if (project.plugins.hasPlugin("org.jetbrains.kotlin.multiplatform")) {
+        // needed to avoid the mpp plugin writing over our artifact ids
+        project.afterEvaluate { project ->
+          it.artifactId = artifactId.forMultiplatform(it, project)
+        }
+      } else {
+        it.artifactId = artifactId
+      }
```
gabrielittner, line 176: Restructured this a bit. The afterEvaluate is only done for multiplatform and in this case we only set it there instead of setting it regularly and then in afterEvaluate.

vanniktech (approved):

vanniktech (approved):

eygraber:
> We don't set project.group and project.version anymore

I'm currently using `VERSION_NAME` property in all of my projects instead of `VERSION`. With this change should I switch to using `VERSION`? 

I don't think I was using `VERSION_NAME` for any reason other than the docs used to say to do that.

gabrielittner:
There is no real need to change anything. If you prefer using the DSL for configuration over properties you can switch from `VERSION_NAME` to the new `coordinates` methods. But  `VERSION_NAME` is still supported and will continue to ve supported.

The reason why we don't set project.version anymore is just for consistency and afaik there is no real need for it since we just set the version directly on the publication.

eygraber:
The only reason I'd want to change it is I don't know if I've been relying on having the project's version set by it. I haven't explicitly been relying on it, but who knows if some other tool might be.

I guess I'd also want to change it in case anyone ever questions what it is in the future and there's no documentation on it anymore.

gabrielittner:
> The only reason I'd want to change it is I don't know if I've been relying on having the project's version set by it. I haven't explicitly been relying on it, but who knows if some other tool might be.

That's a valid reason. We will still use `project.version` as default so you can replace the property with it. You could also do `project.version = project.property("VERSION_NAME")` which is effectively what this plugin did before.

> I guess I'd also want to change it in case anyone ever questions what it is in the future and there's no documentation on it anymore.

It is documented on the following 2 pages if you select `gradle.properties`:
https://vanniktech.github.io/gradle-maven-publish-plugin/central/#configuring-the-pom
https://vanniktech.github.io/gradle-maven-publish-plugin/other/#configuring-the-pom

eygraber:
Do I lose anything by using `VERSION` instead of `VERSION_NAME`, or are they functionally the same (assuming I don't need a different `project.version` and publication version)?

> It is documented

Nice website, didn't see it before

gabrielittner:
You're not loosing anything. `project.version` is the default value for the `version` property that we use https://github.com/vanniktech/gradle-maven-publish-plugin/blob/master/plugin/src/main/kotlin/com/vanniktech/maven/publish/MavenPublishBaseExtension.kt#L28-L29 and if `VERSION_NAME` is set the value of that would then simply override that value in our internal property.

PR #474, opened 2022-12-23
make it possible to publish platforms and version catalogs

### LINKED ISSUE #210: Consider adding support for publishing "platform" metadata
Gradle [supports](https://docs.gradle.org/current/userguide/platforms.html) publishing something like Maven BOMs, called "platforms". Publishing them requires adding the `javaPlatform` component to the publication, as described [here](https://docs.gradle.org/current/userguide/java_platform_plugin.html#sec:java_platform_publishing).

gabrielittner:
Closes #210


plugin/src/main/kotlin/com/vanniktech/maven/publish/Platform.kt
```diff
@@ -294,14 +294,72 @@ data class KotlinJs @JvmOverloads constructor(
...
+      it.from(project.components.getByName("javaPlatform"))
+    }
+  }
+
+  override fun equals(other: Any?): Boolean = other is JavaPlatform
+  override fun hashCode(): Int = this::class.hashCode()
+}
+
+/**
+ * To be used for `org.jetbrains.kotlin.js` projects. Applying this creates a publication for the component called
```
vanniktech, line 335: Is `org.jetbrains.kotlin.js` a copy paste error?

vanniktech (approved):

plugin/src/main/kotlin/com/vanniktech/maven/publish/Platform.kt
gabrielittner: Yes
```suggestion
 * To be used for `version-catalog` projects. Applying this creates a publication for the component called
```

vanniktech (approved):

PR #476, opened 2022-12-23
set User-Agent header for requests to sonatype

### LINKED ISSUE #471: Explicitly set user agent in OkHttp when creating staging repository
Currently the default `okhttp/4.10` is used. Probably should be `pluginId/pluginVersion`.

gabrielittner:
Closes #471


plugin/src/main/kotlin/com/vanniktech/maven/publish/sonatype/SonatypeRepositoryBuildService.kt
```diff
@@ -30,6 +31,8 @@ internal abstract class SonatypeRepositoryBuildService : BuildService<SonatypeRe
       baseUrl = parameters.sonatypeHost.get().apiBaseUrl(),
       username = parameters.repositoryUsername.get(),
       password = parameters.repositoryPassword.get(),
+      userAgentName = BuildConfig.NAME,
+      userAgentVersion = BuildConfig.VERSION
```
vanniktech, line 35:
```suggestion
      userAgentVersion = BuildConfig.VERSION,
```

vanniktech (approved):

PR #477, opened 2022-12-26
run all tests also with the base plugin

gabrielittner:
Closes the biggest gap in our test coverage and moves the last 2 old tests which were using the base plugin to our new setup. The tests are executed in their own shard in parallel to the others.

Thanks to the new tests I also found out that the Kotlin/JS config needs to be done in `afterEvaluate`, the js plugin seems to create the software component lazily. This was hidden for the main plugin because the config there already runs in afterEvaluate.


vanniktech (approved):
Superb

PR #479, opened 2022-12-26
set User-Agent header for requests to sonatype, second attempt 

gabrielittner:
I did revert it because it was breaking snapshots in a really weird way: https://github.com/vanniktech/gradle-maven-publish-plugin/actions/runs/3768895459

Turned out to be a Gradle classpath issue and adding a dependency to build-logic works around that https://github.com/gradle/gradle/issues/17559


gradle/libs.versions.toml
```diff
@@ -32,3 +33,7 @@ truth = { module = "com.google.truth:truth", version.ref = "truth" }
 truth-java8 = { module = "com.google.truth.extensions:truth-java8-extension", version.ref = "truth" }
 truth-testKit = "com.autonomousapps:testkit-truth:1.1"
 maven-model = "org.apache.maven:maven-model:3.8.6"
+
+[plugins]
+
```
vanniktech, line 38:
```suggestion
```

vanniktech (approved):

PR #486, opened 2022-12-29
also setup test fixtures sources jar for Kotlin/JVM, suppress warnings

gabrielittner:
Forgot that Kotlin/JVM projects are configured separately from java/java-library when I added the workaround to publish sources for test fixtures. It is now applied to both of them.

Also added supressions for the following warnings that always appear when publishing a project with test fixtures, regardless of the sources. They happen because there is no concept like a test fixture dependency in Maven so it can't be reflected in the POM.
```
Maven publication 'maven' pom metadata warnings (silence with 'suppressPomMetadataWarningsFor(variant)'):
  - Variant testFixturesApiElements:
      - Declares capability :eithernet-test-fixtures:unspecified which cannot be mapped to Maven
  - Variant testFixturesRuntimeElements:
      - Declares capability :eithernet-test-fixtures:unspecified which cannot be mapped to Maven
These issues indicate information that is lost in the published 'pom' metadata file, which may be an issue if the published library is consumed by an old Gradle version or Apache Maven.
The 'module' metadata file, which is used by Gradle 6+ is not affected.
```

### DIFF
--- a/plugin/src/main/kotlin/com/vanniktech/maven/publish/Platform.kt
+++ b/plugin/src/main/kotlin/com/vanniktech/maven/publish/Platform.kt
@@ -439,7 +439,6 @@ private fun setupTestFixtures(project: Project, sourcesJar: Boolean) {
       it.suppressPomMetadataWarningsFor("testFixturesApiElements")
       it.suppressPomMetadataWarningsFor("testFixturesRuntimeElements")
       it.suppressPomMetadataWarningsFor("testFixturesSourcesElements")
-
     }
   }
 }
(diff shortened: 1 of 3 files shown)


plugin/src/main/kotlin/com/vanniktech/maven/publish/Platform.kt
```diff
@@ -430,6 +425,25 @@ private fun MavenPublication.withJavadocJar(factory: () -> TaskProvider<*>?) {
...
+      project.serviceOf<JvmModelingServices>().createJvmVariant("testFixtures") {
+        it.withSourcesJar().published()
+      }
+    }
+
+    // test fixtures can't be mapped to the POM because there is no equivalent concept in Maven
+    project.mavenPublications {
+      it.suppressPomMetadataWarningsFor("testFixturesApiElements")
+      it.suppressPomMetadataWarningsFor("testFixturesRuntimeElements")
+      it.suppressPomMetadataWarningsFor("testFixturesSourcesElements")
```
gabrielittner, line 441: This last one isn't present in eithernet because it gets added by the `it.withSourcesJar().published()` above which didn't run on eithernet since it's a Kotlin project

vanniktech (approved):

plugin/src/main/kotlin/com/vanniktech/maven/publish/Platform.kt
sschuberth: BTW, is there currently a convenient way to do the same for custom feature variants? I'm having trouble finding a way to access `mavenPublications` in my project using `mavenPublishing`.

PR #490, opened 2023-01-02
add workaround for test fixtures issue

gabrielittner:
This started happening now because of the change to not set project.group and version anymore. As a workaround we will start doing that again but only when the test-fixtures plugin is applied. Unfortunately there doesn't seem to be a way to modify capabilities. The only other way would be to rewrite the generated metadata file on our own but that seems a lot more error prone to me.

@ZacSweers already tested this manually in eithernet https://github.com/slackhq/EitherNet/commit/27270d75ae9bdb1e7b1479ba9ef87f30f66a32a1

Filed Gradle issue https://github.com/gradle/gradle/issues/23354


ZacSweers (approved):

gabrielittner:
Needed to change the code a bit and use `afterEvaluate` + `get()`. Usually when things in Gradle have `Object` as type it will unwrap providers but for group and version it just does `toString()` [28449df](https://github.com/vanniktech/gradle-maven-publish-plugin/pull/490/commits/28449df76a35d352832d37954db96fc6e343437b)

vanniktech (commented):
👍

PR #491, opened 2023-01-02
automate more parts of the publishing process

gabrielittner:
Automates most of the release steps, especially the Github Release creation which was easy to forget. The remaining manual steps are:
1. update changelog, we could probably also have some automation here but I personally don't like these auto generated changelogs
2. create and push a tag for the new version

To make this possible there are a few changes:
- slightly updated the `CHANGELOG.md` format
- the release workflow will extract the section for the new release and automatically create a Github release with it
- `VERSION_NAME` is not comitted anymore and instead always computed
- for releases we tag the tag name as version
- for snapshots we take the last tag increment the last number by 1 and then append SNAPSHOT
- for the website we automatically replace the `<latest-version>` placeholder with the last tag
- snapshot and website workflows run after the release so that there is a snapshot for the next release and the website contains the latest version

Already tested this [here](http://github.com/gabrielittner/renderer)


.github/workflows/publish-snapshot.yml
```diff
@@ -14,22 +16,25 @@ jobs:
...
+      - name: Set version
         run: |
-          echo "VERSION_NAME=$(cat gradle.properties | grep -w "VERSION_NAME" | cut -d'=' -f2)" >> $GITHUB_ENV
+          echo "ORG_GRADLE_PROJECT_VERSION_NAME=$(git describe --tags --abbrev=0 | awk -F. '/[0-9]+\./{$NF++;print}' OFS=.)-SNAPSHOT" >> $GITHUB_ENV
+
+      - uses: gradle/gradle-build-action@v2
 
-      - name: Publish snapshot
-        run: ./gradlew publishAllPublicationsToMavenCentralRepository
-        if: endsWith(env.VERSION_NAME, '-SNAPSHOT')
```
vanniktech, line 32: Should not this stay?

.github/workflows/publish-snapshot.yml
```diff
@@ -4,6 +4,8 @@ on:
   push:
     branches:
       - main
+    tags:
```
vanniktech, line 7: We don't need this here do we?

vanniktech (approved):

.github/workflows/publish-snapshot.yml
gabrielittner: It's not strictly needed. I added it so that when a new release created we also create the first snapshot for the release after that. Kind of mirrors the behavior that we had before from the "Prepare next development version" commit which would also cause a snapshot to be published

.github/workflows/publish-snapshot.yml
gabrielittner: The `if` was only needed because the release commit would contain a non snapshot version which would then make this workflow fail when it runs on that commit. Since that commit doesn't exist anymore and we are always computing a snapshot version we can just always run it

.github/workflows/publish-snapshot.yml
vanniktech: Nice!

.github/workflows/publish-snapshot.yml
vanniktech: Got it

Issue #496, opened 2023-01-15
Error in publish when profile name does not matches with library group name


PatilShreyas:
## Background

When I created an account at oss.sonatype.org, I did a mistake in my profile name and it was permanently set to ***`dev.shreyaspatil.android`***. But I always publish my libraries with the following pattern: `dev.shreyaspatil.LIB_NAME:ARTIFACT:VERSION`. i.e. I don't use ***".android"*** in the group.  See [this discussion](https://issues.sonatype.org/browse/OSSRH-64470?focusedCommentId=1016618&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-1016618)

## Issue

When I'm using version "0.23.1" of this plugin and hitting `./gradlew publish`, it's failing in the task `createStagingRepository` with the following error:

```
No matching staging profile found in account patilshreyas. It is expected that the account contains a staging profile that matches or is the start of dev.shreyaspatil.compose-compiler-report-generator.Available profiles are: dev.shreyaspatil.android
```

**Then I checked source of the plugin and found this below code:**

https://github.com/vanniktech/gradle-maven-publish-plugin/blob/6686704e1dcd29a80325c65f6dcad01fad19cb9b/nexus/src/main/kotlin/com/vanniktech/maven/publish/nexus/Nexus.kt#L51-L59

Here, the issue is actually a calculation of `candidateProfiles` since it only takes values from `allProfiles` that matches `group == it.name || group.startsWith(it.name)`. Due to my case (_that I explained in the background section_), plugin throws the above mentioned error and I'm unable to publish artifacts.

I hope this information is helpful. Let me know if any more info is required. Would be happy to contribute to fixing this as well.

gabrielittner:
Thanks for the detailed report. I think we can at least add a fallback that if there is just one staging profile we will use that.

PatilShreyas:
Raised a PR for fix: #497

gabrielittner:
I'm planning to publish 0.23.2 with the fix next week. There is already a `0.23.2-SNAPSHOT` in the mean time

PatilShreyas:
Thanks @gabrielittner. It's very quick!

PR #497, opened 2023-01-15
Select single profile if only one profile is available else filter them for group

PatilShreyas:
Fixes: #496


gabrielittner:
Thank you for sending the pr. I missed it before I opened mine (#498) in which I split up the logic a bit more and also added a shared prefix case.

PatilShreyas:
> Thank you for sending the pr. I missed it before I opened mine (#498) in which I split up the logic a bit more and also added a shared prefix case.

Oh great!

PR #498, opened 2023-01-15
make finding a staging profile more lenient

### LINKED ISSUE #496: Error in publish when profile name does not matches with library group name

gabrielittner:
Closes #496 

The logic is now:
- if there is just one profile always use that regardless of whether the name matches
- profile that equals the group id
- profile that is a prefix for the group id
- profile that shares a prefix with the group id

### DIFF
--- a/nexus/src/main/kotlin/com/vanniktech/maven/publish/nexus/Nexus.kt
+++ b/nexus/src/main/kotlin/com/vanniktech/maven/publish/nexus/Nexus.kt
@@ -48,12 +48,24 @@ class Nexus(
       throw IllegalArgumentException("No staging profiles found in account $username. Make sure you called \"./gradlew publish\".")
     }
 
-    val candidateProfiles = allProfiles.filter { group == it.name || group.startsWith(it.name) }
+    if (allProfiles.size == 1) {
+      return allProfiles[0]
+    }
+
+    var candidateProfiles = allProfiles.filter { group == it.name }
+
+    if (candidateProfiles.isEmpty()) {
+      candidateProfiles = allProfiles.filter { group.startsWith(it.name) }
+    }
+
+    if (candidateProfiles.isEmpty()) {
+      candidateProfiles = allProfiles.filter { group.commonPrefixWith(it.name).isNotEmpty() }
+    }
 
     if (candidateProfiles.isEmpty()) {
       throw IllegalArgumentException(
         "No matching staging profile found in account $username. It is expected that the account contains a staging " +
-          "profile that matches or is the start of $group." +
+          "profile that matches or is the start of $group. " +
           "Available profiles are: ${allProfiles.joinToString(separator = ", ") { it.name }}"
       )
     }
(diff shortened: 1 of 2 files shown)


nexus/src/main/kotlin/com/vanniktech/maven/publish/nexus/Nexus.kt
```diff
@@ -48,12 +48,24 @@ class Nexus(
...
+
+    var candidateProfiles = allProfiles.filter { group == it.name }
+
+    if (candidateProfiles.isEmpty()) {
+      candidateProfiles = allProfiles.filter { group.startsWith(it.name) }
+    }
+
+    if (candidateProfiles.isEmpty()) {
+      candidateProfiles = allProfiles.filter { group.commonPrefixWith(it.name).isNotEmpty() }
+    }
```
PatilShreyas, line 63: This can be done in one-liner as...

```kotlin
val candidateProfiles = allProfiles.filter { group == it.name }
    .ifEmpty { allProfiles.filter { group.startsWith(it.name) } }
    .ifEmpty { allProfiles.filter { group.commonPrefixWith(it.name).isNotEmpty() } }
```

gabrielittner: 👍 updated it [e2a725f](https://github.com/vanniktech/gradle-maven-publish-plugin/pull/498/commits/e2a725f81737b9506d9f5e29fda774f4060fe21e)

vanniktech (approved):

PR #500, opened 2023-01-17
test against AGP 8.0.0-beta01 and 8.1.0-alpha01

gabrielittner:


gabrielittner:
Not merging this for now, looks like both releases are incompatible with Gradle 8.0.0-rc-2

vanniktech (approved):

PR #504, opened 2023-01-26
Open up SonatypeHost

### LINKED ISSUE #503: How to set a custom host in publishToMavenCentral?
```
fun publishToMavenCentral(host: SonatypeHost = SonatypeHost.DEFAULT, automaticRelease: Boolean = false) {
```

SonatypeHost is an enum, so how to set a custom host in the form of xxx.oss.sonatype.org for publishing to Sonetype?

DEFAULT and use SONATYPE_HOST env? Can the enum be opened up to a value type?

yschimke:
Relates to https://issues.sonatype.org/browse/OSSRH-88192

Make SonatypeHost a value class instead of an enum.

fixes #503

### DIFF
--- a/plugin/src/main/kotlin/com/vanniktech/maven/publish/MavenPublishBaseExtension.kt
+++ b/plugin/src/main/kotlin/com/vanniktech/maven/publish/MavenPublishBaseExtension.kt
@@ -81,7 +81,8 @@ abstract class MavenPublishBaseExtension(
   }
 
   @JvmOverloads
-  fun publishToMavenCentral(host: String = "DEFAULT", automaticRelease: Boolean = false) {
+  @JvmSynthetic
+  fun publishToMavenCentral(host: String, automaticRelease: Boolean = false) {
     publishToMavenCentral(SonatypeHost.valueOf(host), automaticRelease)
   }
 
--- a/plugin/src/main/kotlin/com/vanniktech/maven/publish/MavenPublishBaseExtension.kt
+++ b/plugin/src/main/kotlin/com/vanniktech/maven/publish/MavenPublishBaseExtension.kt
@@ -80,6 +80,11 @@ abstract class MavenPublishBaseExtension(
     project.tasks.registerDropRepository(buildService)
   }
 
+  @JvmOverloads
+  fun publishToMavenCentral(host: String = "DEFAULT", automaticRelease: Boolean = false) {
+    publishToMavenCentral(SonatypeHost.valueOf(host), automaticRelease)
+  }
+
   /**
    * Automatically apply Gradle's `signing` plugin and configure all publications to be signed. If signing credentials
    * are not configured this will fail the build unless the current version is a `SNAPSHOT`.
--- a/plugin/src/main/kotlin/com/vanniktech/maven/publish/SonatypeHost.kt
+++ b/plugin/src/main/kotlin/com/vanniktech/maven/publish/SonatypeHost.kt
@@ -8,12 +8,9 @@ import org.gradle.api.provider.Provider
  *
  * https://central.sonatype.org/articles/2021/Feb/23/new-users-on-s01osssonatypeorg/
  */
-enum class SonatypeHost(
+class SonatypeHost(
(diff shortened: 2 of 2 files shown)


yschimke:
Testing failed

`org.gradle.internal.snapshot.ValueSnapshottingException: Could not serialize value of type SonatypeHost`

yschimke:
Now working at least to the point of a 403.

```
Caused by: java.io.IOException: Cannot get stagingProfiles for account yuri@schimke.ee: <html>
  <head>
    <title>403 - Forbidden</title>
    <meta http-equiv="Content-Type" content="text/html; charset=utf-8"/>

    <link rel="icon" type="image/png" href="https://google.oss.sonatype.org/favicon.png">
    <!--[if IE]>
    <link rel="SHORTCUT ICON" href="https://google.oss.sonatype.org/favicon.ico"/>
    <![endif]-->
```

plugin/src/main/kotlin/com/vanniktech/maven/publish/SonatypeHost.kt
```diff
@@ -28,4 +26,15 @@ enum class SonatypeHost(
...
+
+  companion object {
+    fun valueOf(sonatypeHost: String): SonatypeHost = when (sonatypeHost) {
+      "DEFAULT" -> DEFAULT
+      "S01" -> S01
+      else -> throw IllegalArgumentException("No SonatypeHost constant $sonatypeHost")
+    }
+
+    val DEFAULT = SonatypeHost("https://oss.sonatype.org")
+    val S01 = SonatypeHost("https://s01.oss.sonatype.org")
```
gabrielittner, line 38: Could you add `@JvmStatic` and `@JvmField` to the function and properties respectively? That should keep compatibility for Groovy users.

gabrielittner (commented):
I think we also need an overload like this in `MavenPublishBaseExtension`: 
```
@JvmOverloads
fun publishToMavenCentral(host: String = "DEFAULT" automaticRelease: Boolean = false) 
```
Groovy (or maybe just Gradle) has something that allows passing an enum value as string to a function.

yschimke:
> I think we also need an overload like this in `MavenPublishBaseExtension`: 
> ```
> @JvmOverloads
> fun publishToMavenCentral(host: String = "DEFAULT" automaticRelease: Boolean = false) 
> ```
> Groovy (or maybe just Gradle) has something that allows passing an enum value as string to a function.

I'm not confident about this part of the change. Won't the overload conflict when no params are provided? Like `publishToMavenCentral()`

yschimke:
Feel free to takeover this PR, or suggest further changes.

plugin/src/main/kotlin/com/vanniktech/maven/publish/MavenPublishBaseExtension.kt
```diff
@@ -80,6 +80,11 @@ abstract class MavenPublishBaseExtension(
     project.tasks.registerDropRepository(buildService)
   }
 
+  @JvmOverloads
+  fun publishToMavenCentral(host: String = "DEFAULT", automaticRelease: Boolean = false) {
```
yschimke, line 84: Should I use @ JvmSynthetic to hide from Kotlin?

gabrielittner: That's a good idea.

We can also leave out the default value for `host` to avoid your concern about conflicting overloads

gabrielittner:
Looks like there is a ktlint issue, can you run `ktlintFormat` once? Otherwise I can also do that before merging
```
> Task :plugin:ktlintMainSourceSetCheck FAILED
/home/runner/work/gradle-maven-publish-plugin/gradle-maven-publish-plugin/plugin/src/main/kotlin/com/vanniktech/maven/publish/SonatypeHost.kt:14:2 Missing spacing before ":"
```

gabrielittner (approved):
Thanks for the contribution. I can create a new release with this soon, but we're also blocked by the Sonatype issues 😬

PR #526, opened 2023-02-19
run tests with config cache enabled

gabrielittner:
The publishing plugin is already supporting configuration caching since Gradle 7.6 and in 8.1 the signing plugin is added. With this our tests will run with config caching enabled on those version depending on whether the test uses signing or not. For Gradle 8.1 the tests run against a nightly.


vanniktech (approved):

plugin/src/integrationTest/kotlin/com/vanniktech/maven/publish/ProjectSpecRunner.kt
```diff
@@ -18,11 +20,17 @@ fun ProjectSpec.run(fixtures: Path, temp: Path, options: TestOptions): ProjectRe
...
+    arguments.add("--configuration-cache")
+  }
+
   val result = GradleRunner.create()
     .withGradleVersion(options.gradleVersion.value)
     .withProjectDir(project.toFile())
     .withDebug(true)
-    .withArguments(task, "--stacktrace")
+    .withArguments(arguments)
+    .forwardOutput()
```
gabrielittner, line 33:
```suggestion
```

PR #545, opened 2023-03-23
fix snapshot publishing

### LINKED ISSUE #544: Updating to 0.25.0 fails when publishing SNAPSHOTs
I have a few projects where all I did was update from 0.24.0 to 0.25.0 and I'm not getting failures when building SNAPSHOTs:

```
Execution failed for task ':ejson:publishMavenPublicationToMavenCentralRepository'.
> The staging repository was not created yet. Please open a bug with a build scan or build logs and stacktrace
```

Repos:
 - https://github.com/eygraber/ejson-kotlin/actions/runs/4504866394
 - https://github.com/eygraber/compose-color-picker/actions/runs/4504863851

gabrielittner:
The `stagingRepositoryId.isPresent` check in `publishingUrl` will cause the `Provider` to be evaluated which then will make `requireNotNull` throw because `stagingRepositoryId` is expected to be `null` in this case. I moved the check into `publishingUrl` to resolve this. 

Tested the solution manually by publishing to maven local and trying out both snapshot and release publishing in another project using that version. I need to find a way to have this as an automated test somehow.

Closes #544

### DIFF
--- a/plugin/src/main/kotlin/com/vanniktech/maven/publish/ProjectExtensions.kt
+++ b/plugin/src/main/kotlin/com/vanniktech/maven/publish/ProjectExtensions.kt
@@ -7,6 +7,7 @@ import org.gradle.api.Project
 import org.gradle.api.publish.PublishingExtension
 import org.gradle.api.publish.maven.MavenPublication
 import org.gradle.plugins.signing.SigningExtension
+import org.gradle.util.GradleVersion
 import org.jetbrains.kotlin.gradle.plugin.KotlinBasePlugin
 
 internal fun Project.findOptionalProperty(propertyName: String) = findProperty(propertyName)?.toString()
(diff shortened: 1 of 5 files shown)


plugin/src/main/kotlin/com/vanniktech/maven/publish/ProjectExtensions.kt
```diff
@@ -58,3 +59,11 @@ internal fun Project.isAtLeastUsingAndroidGradleVersion(major: Int, minor: Int,
     false
   }
 }
+
+@Suppress("UnstableApiUsage")
+internal fun Project.configurationCache(): Boolean {
+  if (GradleVersion.current() > GradleVersion.version("7.6")) {
```
gabrielittner, line 65:
```suggestion
  if (GradleVersion.current() >= GradleVersion.version("7.6")) {
```

vanniktech (approved):

PR #554, opened 2023-04-12
Gradle 8.1

gabrielittner:


plugin/src/main/kotlin/com/vanniktech/maven/publish/Platform.kt
```diff
@@ -493,13 +492,13 @@ private fun setupTestFixtures(project: Project, sourcesJar: Boolean) {
...
         val extension = project.extensions.getByType(JavaPluginExtension::class.java)
         val testFixturesSourceSet = extension.sourceSets.maybeCreate(variant)
-        val method = services.javaClass.getMethod("createJvmVariant", String::class.java, SourceSet::class.java, Action::class.java)
-        method.invoke(services, variant, testFixturesSourceSet, action)
+        services.createJvmVariant(variant, testFixturesSourceSet, action)
       } else {
-        project.serviceOf<JvmModelingServices>().createJvmVariant(variant, action)
+        val method = services.javaClass.getMethod("createJvmVariant", String::class.java, Action::class.java)
+        method.invoke(services, variant, action)
       }
```
gabrielittner, line 502: Just swapped which version does the direct method call and which uses reflection

vanniktech (approved):

Issue #573, opened 2023-05-03
Are the docs for local repos actual? Should myRepo be replaced with mavenLocal?


andrew-k-21-12:
Hello-hello 🙂 

I am configuring publishing to some local Maven repository according to the instructions at:
https://vanniktech.github.io/gradle-maven-publish-plugin/other/#applying-the-plugin

The actual result is that declarations like:
```kotlin
publishing {
    repositories {
        myRepo {
            url = uri(layout.buildDirectory.dir("repo"))
        }
    }
}
```

lead to:
> Unresolved reference: myRepo

At the same time the following construction works nicely:
```kotlin
publishing {
    repositories {
        mavenLocal {
            url = uri(layout.buildDirectory.dir("repo"))
        }
    }
}
```

Ready to provide a sample project or additional details if needed.

PR #575, opened 2023-05-05
Align the repo name between kotlin and groovy

BraisGabin:


gabrielittner:
Thank you. Based on #573 it turns out that the syntax was generally wrong so I also updated that in here

PR #592, opened 2023-06-22
Update tested versions

gabrielittner:
Changes based on the new versions:
- Gradle completely changed the API used to model test fixtures (`JvmModelingServices` and `JvmVariantBuilder` don't exist at all anymore)
    - ~I haven't found a new workaround yet since the replacement is just an [instantiated class](https://github.com/gradle/gradle/blob/master/subprojects/plugins/src/main/java/org/gradle/api/plugins/JavaTestFixturesPlugin.java#L67-L74) and I didn't find a way to obtain it. So for now we don't apply a workaround to enable test fixture source publishing on 8.2+.~
    - Added a new workaround that works from 8.1 on, so the workaround only has 2 variants and not 3
    - Adjusted the existing workaround to not directly reference any of the removed types which will make it possible to compile against 8.2 when it is out
- Kotlin/JS adds an additional dependency by default which our tests need to account for
- Kotlin MPP now supports config caching so it's now enabled in multiplatform tests

### DIFF
--- a/plugin/src/integrationTest/kotlin/com/vanniktech/maven/publish/ProjectSpecRunner.kt
+++ b/plugin/src/integrationTest/kotlin/com/vanniktech/maven/publish/ProjectSpecRunner.kt
@@ -40,12 +40,13 @@ fun ProjectSpec.run(fixtures: Path, temp: Path, options: TestOptions): ProjectRe
 }
 
 private fun TestOptions.supportsConfigCaching(plugins: List<PluginSpec>): Boolean {
-  // TODO: Kotlin Multiplatform plugin has configuration cache issues
-  //  - https://youtrack.jetbrains.com/issue/KT-49933 (meta ticket)
-  //  - https://youtrack.jetbrains.com/issue/KT-43293 (closed but still has issues)
-  //  - https://youtrack.jetbrains.com/issue/KT-55051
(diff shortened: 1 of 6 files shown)


plugin/src/integrationTest/kotlin/com/vanniktech/maven/publish/ProjectSpecRunner.kt
```diff
@@ -40,12 +40,13 @@ fun ProjectSpec.run(fixtures: Path, temp: Path, options: TestOptions): ProjectRe
 }
 
 private fun TestOptions.supportsConfigCaching(plugins: List<PluginSpec>): Boolean {
-  // TODO: Kotlin Multiplatform plugin has configuration cache issues
-  //  - https://youtrack.jetbrains.com/issue/KT-49933 (meta ticket)
-  //  - https://youtrack.jetbrains.com/issue/KT-43293 (closed but still has issues)
-  //  - https://youtrack.jetbrains.com/issue/KT-55051
-  if (plugins.any { it.id == kotlinMultiplatformPlugin.id }) {
-    return false
+  // Kotlin MPP support s config cache since 1.9.0
```
vanniktech, line 43:
```suggestion
  // Kotlin MPP supports config cache since 1.9.0
```

vanniktech (approved):

PR #677, opened 2023-12-07
fix: resolve issue with attempting to register a task by the same name more than once.

autonomousapps:
I wasn't sure what the best resolution was to the problem I faced, so I figured it would be best to just show a PR and we could discuss. This is the error I ran into in my (closed source) project:
```
Caused by: org.gradle.api.internal.tasks.DefaultTaskContainer$DuplicateTaskException: Cannot add task 'simpleJavadocJar' as a task with that name already exists.
at org.gradle.api.internal.tasks.DefaultTaskContainer.failOnDuplicateTask(DefaultTaskContainer.java:257)	
•••
at com.vanniktech.maven.publish.tasks.JavadocJar$Companion.plainJavadocJar(JavadocJar.kt:27)	
at com.vanniktech.maven.publish.tasks.JavadocJar$Companion.javadocJarTask$plugin(JavadocJar.kt:19)	
at com.vanniktech.maven.publish.GradlePlugin$configure$1$1.invoke(Platform.kt:85)	
at com.vanniktech.maven.publish.GradlePlugin$configure$1$1.invoke(Platform.kt:85)	
at com.vanniktech.maven.publish.PlatformKt.withJavadocJar(Platform.kt:444)	
at com.vanniktech.maven.publish.PlatformKt.access$withJavadocJar(Platform.kt:1)	
at com.vanniktech.maven.publish.GradlePlugin.configure$lambda$0(Platform.kt:85)	
at com.vanniktech.maven.publish.ProjectExtensionsKt.mavenPublicationsWithoutPluginMarker$lambda$0(ProjectExtensions.kt:32)
```
I did a bit of tracing and it looks like you're iterating over every publication in a project, and then adding one of the tasks for each. I have a plugin to support testing gradle plugins that also [adds a publication for test purposes](https://github.com/autonomousapps/dependency-analysis-gradle-plugin/blob/main/testkit/gradle-testkit-plugin/src/main/kotlin/com/autonomousapps/GradleTestKitSupportExtension.kt#L173-L187).

### DIFF
--- a/plugin/src/main/kotlin/com/vanniktech/maven/publish/Platform.kt
+++ b/plugin/src/main/kotlin/com/vanniktech/maven/publish/Platform.kt
@@ -98,9 +98,11 @@ data class GradlePlugin @JvmOverloads constructor(
       "Calling configure(GradlePlugin(...)) requires the java-gradle-plugin to be applied"
     }
 
+    val javadocJarTask = project.javadocJarTask(javadocJar)
+
     project.mavenPublicationsWithoutPluginMarker {
       it.withJavaSourcesJar(sourcesJar, project)
-      it.withJavadocJar { project.javadocJarTask(javadocJar) }
+      it.withJavadocJar { javadocJarTask }
     }
   }
 }
--- a/plugin/src/main/kotlin/com/vanniktech/maven/publish/tasks/JavadocJar.kt
+++ b/plugin/src/main/kotlin/com/vanniktech/maven/publish/tasks/JavadocJar.kt
@@ -4,7 +4,6 @@ import com.vanniktech.maven.publish.JavadocJar as JavadocJarOption
 import com.vanniktech.maven.publish.JavadocJar.Dokka.DokkaTaskName
 import com.vanniktech.maven.publish.JavadocJar.Dokka.ProviderDokkaTaskName
 import com.vanniktech.maven.publish.JavadocJar.Dokka.StringDokkaTaskName
-import org.gradle.api.InvalidUserDataException
 import org.gradle.api.Project
 import org.gradle.api.tasks.TaskProvider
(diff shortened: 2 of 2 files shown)


gabrielittner:
I think I probably just made the assumption that a java-gradle-plugin project only has the default publications that are created automatically and that there is just one publication that isn't a plugin marker. 

As an alternative we could move the [javadocJarTask call](https://github.com/vanniktech/gradle-maven-publish-plugin/blob/main/plugin/src/main/kotlin/com/vanniktech/maven/publish/Platform.kt#L103) outside of the loop. There isn't a scenario with 0 publications for which we'd then create an unneeded task. [KMP projects](
https://github.com/vanniktech/gradle-maven-publish-plugin/blob/main/plugin/src/main/kotlin/com/vanniktech/maven/publish/Platform.kt#L286-L290) already use that approach.

autonomousapps:
> As an alternative we could move the [javadocJarTask call](https://github.com/vanniktech/gradle-maven-publish-plugin/blob/main/plugin/src/main/kotlin/com/vanniktech/maven/publish/Platform.kt?rgh-link-date=2023-12-08T17%3A46%3A55Z#L103) outside of the loop.

I can give this a try. Thanks for the review!

autonomousapps:
Actually, should we pull this out of the loop for all platforms? I ran into the issue with a gradle plugin, but it might happen in other cases?

If you prefer to see that happen in the real world before making such a change, I understand.

gabrielittner (approved):
Thanks for the fix

> Actually, should we pull this out of the loop for all platforms? I ran into the issue with a gradle plugin, but it might happen in other cases?

If I didn't miss anything the plugin is only looping over publications for Gradle plugins and KMP. For the other platforms the underying plugins don't create publications automatically, so the plugin is creating one manually and only creates the task there.

Issue #624, opened 2023-08-08
Attempting to configure publishing platform object results in build failure


mattinger:
I am attempting to use the plugin to configure an android library that has a single product dimension.  however, when i try to configure the platform in the mavenPublish block, i always get a failure. I've simplified this down to a project that doesn't even have any dimensions to it.  I consistently get this error:

```
Cause 1: java.lang.IllegalStateException: The value for this property is final and cannot be changed any further.
	at org.gradle.api.internal.provider.AbstractProperty$FinalizedValue.beforeMutate(AbstractProperty.java:498)
	at org.gradle.api.internal.provider.AbstractProperty.assertCanMutate(AbstractProperty.java:272)
	at org.gradle.api.internal.provider.AbstractProperty.setSupplier(AbstractProperty.java:221)
	at org.gradle.api.internal.provider.DefaultProperty.set(DefaultProperty.java:71)
	at com.vanniktech.maven.publish.MavenPublishBaseExtension.configure(MavenPublishBaseExtension.kt:346)
	at com.vanniktech.maven.publish.MavenPublishPluginKt$configurePlatform$2$1.invoke(MavenPublishPlugin.kt:45)
	at com.vanniktech.maven.publish.MavenPublishPluginKt$configurePlatform$2$1.invoke(MavenPublishPlugin.kt:42)
	at com.android.build.api.extension.impl.AndroidComponentsExtensionImpl$finalizeDsl$1.execute(AndroidComponentsExtensionImpl.kt:47)
	at com.android.build.api.extension.impl.AndroidComponentsExtensionImpl$finalizeDsl$1.execute(AndroidComponentsExtensionImpl.kt:46)
```

I debugged through this a bit and i found that while it is setting the value i've put in the mavenPublishing block, the plugin itself is trying to do some finalization related to creating the gradle tasks, which is forcible setting the value again.  But since that value has been finalized, it can't do that.  In addition, it's also making an assumption about the platform type when it does this, attempting to set it back to AndroidSingleVariantLibrary.

```
  plugins.withId("com.android.library") {
    // afterEvaluate is too late, but we can't run this synchronously because we shouldn't call the APIs for
    // multiplatform projects that use Android
    androidComponents.finalizeDsl {
      if (!plugins.hasPlugin("org.jetbrains.kotlin.multiplatform")) {
        val variant = project.findOptionalProperty("ANDROID_VARIANT_TO_PUBLISH") ?: "release"
[...]

Issue #565, opened 2023-04-25
Disable javadoc generation


p-dmitriy:
How can I disable javadoc generation and publishing? There is an `AndroidSingleVariantLibrary` class that have `publishJavadocJar` parameter with default value `true`. And single usage of that class does not allow to override this parameter.

Also how I can see there is no way to replace publishing variant `AndroidSingleVariantLibrary` with multiple variants. If I declare multiple variants in android block, single variant still will be added by plugin.

p-dmitriy:
Found that I need to use base plugin. It is not obviously, why did you not made single plugin but configurable?

gabrielittner:
This will properly work without having to use base plugin in the next release

PR #680, opened 2023-12-09
make it possible to use configure with the main plugin

### LINKED ISSUE #675: Getting "The value for this property is final and cannot be changed any further"
I was trying to configure the plugin with [Dokkatoo](https://github.com/adamko-dev/dokkatoo), an alternative to Dokka, like this:

    configure(KotlinJvm(JavadocJar.Dokka("dokkatooGeneratePublicationJavadoc")))

which gives

```
Caused by: java.lang.IllegalStateException: The value for this property is final and cannot be changed any further.
        at org.gradle.api.internal.provider.AbstractProperty$FinalizedValue.beforeMutate(AbstractProperty.java:498)
        at org.gradle.api.internal.provider.AbstractProperty.assertCanMutate(AbstractProperty.java:272)
        at org.gradle.api.internal.provider.AbstractProperty.setSupplier(AbstractProperty.java:221)
        at org.gradle.api.internal.provider.DefaultProperty.set(DefaultProperty.java:71)
        at com.vanniktech.maven.publish.MavenPublishBaseExtension.configure(MavenPublishBaseExtension.kt:346)
        at com.vanniktech.maven.publish.MavenPublishPluginKt$configurePlatform$3.invoke(MavenPublishPlugin.kt:59)
        at com.vanniktech.maven.publish.MavenPublishPluginKt$configurePlatform$3.invoke(MavenPublishPlugin.kt:50)
        at com.vanniktech.maven.publish.MavenPublishPluginKt.configurePlatform$lambda$2(MavenPublishPlugin.kt:50)
```

For the record, I'm getting the same when using "dokkaHtml" instead of "dokkatooGeneratePublicationJavadoc".

I'm using Gradle 8.5.

gabrielittner:
This brings 2 changes:
- It's possible to call configure from a build script in the main plugin which resolves #675, #624 and #565.
- The underlying functionality of automatic configuration is now exposed in the extension through `configureBasedOnAppliedPlugins`. This allows to get the automatic behavior for the base plugin while keeping the rest configurable manually.

The tests run through without an issue, so moving the automatic KMP configuration to `afterEvaluate` doesn't seem to have any issues. In the first place I only moved it out of `afterEvaluate` in #351 so that it runs before the Android configuration in `finalizeDsl`. That's not really needed since we can just do the KMP configuration in `finalizeDsl` as well and then no-op in `afterEvaluate` if the platform was already configured.

### DIFF
--- a/docs/base.md
+++ b/docs/base.md
@@ -1,8 +1,8 @@
 # Base plugin
 
-Starting with version 0.15.0 there is a base plugin. This new plugin has the same capabilities as the main
-plugin but does not configure anything automatically. In the current stage the APIs are still marked with `@Incubating`
-so they might change.
+The base plugin generally provides the same functionality as the main plugin,
+but it doesn't configure anything automatically and instead relies on manual
+configuration.
 
 ## Applying the plugin
 
@@ -26,10 +26,11 @@ Add the plugin to any Gradle project that should be published
 
 ## General configuration
 
+Follow the steps of the [Maven Central](central.md) or [other Maven repositories](other.md) guides.
+Note that the `SONATYPE_HOST`, `SONATYPE_AUTOMATIC_RELEASE` and `RELEASE_SIGNING_ENABLED` are not
+considered by the base plugin and the appropriate DSL methods need to be called.
 
-Follow the steps of the [Maven Central](central.md) or [other Maven repositories](other.md) guides. Note that the
-configuration of where to publish to with Gradle properties won't work in the base plugin. For the pom configuration
-via Gradle properties the following needs to be enabled in the DSL:
+For the pom configuration via Gradle properties the following needs to be enabled in the DSL:
 
 === "build.gradle"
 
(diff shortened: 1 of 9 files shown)


docs/base.md
```diff
@@ -49,367 +50,29 @@ via Gradle properties the following needs to be enabled in the DSL:
...
-The biggest difference between the regular and the base plugin is that it won't setup what to publish by default. The
-base plugin offers various APIs depending on the type of project that should be published
+To define what should be published, the `configure` method in the `mavenPublishing`
+block needs to be called. Check out the [what to publish page](what.md) which
+contains a detailed description of available options for each project type.
 
-### Android Library (multiple variants)
-
-For projects using the `com.android.library` plugin. This will publish all variants of the project (e.g. both
-`debug` and `release`) or a subset of specified variants.
```
gabrielittner, line 58: Old docs for each platform have been moved to `what.md`

docs/what.md
```diff
@@ -0,0 +1,376 @@
...
+- [`org.jetbrains.kotlin.js`](#kotlin-js-library)
+- [`org.jetbrains.kotlin.multiplatform`](#kotlin-multiplatform-library)
+- [`java`](#java-library)
+- [`java-library`](#java-library)
+- [`java-gradle-plugin`](#gradle-plugin)
+- [`com.gradle.plugin-publish`](#gradle-publish-plugin)
+- [`java-platform`](#java-platform)
+- [`version-catalog`](#version-catalog)
+
+## Android Library (multiple variants)
```
gabrielittner, line 16: From here on this is a 1:1 copy of what used to be in `base.md`

plugin/src/integrationTest/kotlin/com/vanniktech/maven/publish/ProjectSpecRunner.kt
```diff
@@ -178,7 +178,6 @@ private fun writeSettingFile(path: Path) {
 
     dependencyResolutionManagement {
         repositories {
-            mavenLocal()
```
gabrielittner, line 181: Having this made Kotlin/JS compilation fail for some reason https://youtrack.jetbrains.com/issue/KT-61760, fortunately we don't need it.

plugin/src/main/kotlin/com/vanniktech/maven/publish/MavenPublishBaseExtension.kt
```diff
@@ -349,4 +352,66 @@ abstract class MavenPublishBaseExtension(
 
     platform.configure(project)
   }
+
+  /**
+   * Calls [configure] with a [Platform] chosen based on other applied Gradle plugins.
+   */
+  @Incubating
+  fun configureBasedOnAppliedPlugins() {
```
gabrielittner, line 360: We could allow optionally passing `javadocJar: JavadocJar? = null, sourcedJar: Boolean = true` (null = we use our current logic) here to eliminate many use cases of calling `configure`.  Only weird thing about this would be that some platforms would ignore the option if it doesn't apply or is not configurable for it.

vanniktech (approved):

PR #689, opened 2023-12-19
refactor: Make `coordinates()` take `Any` as a `version`

sschuberth:
This allows to conveniently pass Gradle's `Project.version` which is a generic `Object`, see [1].

[1]: https://docs.gradle.org/current/dsl/org.gradle.api.Project.html#org.gradle.api.Project:version

### DIFF
--- a/docs/central.md
+++ b/docs/central.md
@@ -257,7 +257,7 @@ Snapshots can be published by setting the version to something ending with `-SNA
 and then running the following Gradle task:
 
 ```
-./gradlew publishAllPublicationsToMavenCentral
+./gradlew publishAllPublicationsToMavenCentralRepository
 ```
 
 The snapshot will be automatically available in Sonatype's
@@ -282,7 +282,7 @@ The publishing process for Maven Central consists of several steps
 By running the following Gradle task the plugin will take care of steps 1 to 3 automatically:
 
 ```
-./gradlew publishAllPublicationsToMavenCentral --no-configuration-cache
+./gradlew publishAllPublicationsToMavenCentralRepository --no-configuration-cache
 ```
 
 The releasing step can be done manually by going to oss.sonatype.org (or s01.oss.sonatype.org) and


plugin/src/main/kotlin/com/vanniktech/maven/publish/MavenPublishBaseExtension.kt
```diff
@@ -166,7 +166,7 @@ abstract class MavenPublishBaseExtension(
    * artifactIds like `[artifactId]-jvm`.
    */
   @Incubating
-  fun coordinates(groupId: String, artifactId: String, version: String) {
+  fun coordinates(groupId: String, artifactId: String, version: Any) {
```
gabrielittner, line 169: Thanks for the PR. What do you think about having an overload without `version` instead? The property will already default to it if it's not set (line 33) and Gradle will eventually (probably in v9) use a `Provider` for version so having `Any` here goes into the wrong direction when thinking about the future.

gabrielittner: The artifactId will default to the project name or to what the plugin that creates a publication uses. For example Kotlin multiplatform will use `{project.name}-{platform}`. So `artifactId` does not need to be called if the project name already has what you need.

Similar to the others `groupId` will default to `project.getGroup()`

sschuberth: Ok, in that case, let me just drop the `version`-related commit, and limit this PR to the docs fix.

gabrielittner:
Thanks

plugin/src/main/kotlin/com/vanniktech/maven/publish/MavenPublishBaseExtension.kt
sschuberth: Oh, I just realized that `groupId()` (as well as `version()` and `artifactId()`) are private, so I cannot call them directly. From that perspective, it could again make sense to have a `coordinates(group, artifactId)` overload. But then you'd probably want a `coordinates(group)` overload, too. Which leaves us with the question whether it's not better to simply make the underlying functions public instead of having `coordinates()` plus overloads.

However, changing the signature to

    fun coordinates(groupId: String? = null, artifactId: String? = null, version: String? = null)

could make sense, so users set only the needed parts in a single call (with named parameters).

sschuberth: Like this: https://github.com/vanniktech/gradle-maven-publish-plugin/pull/691

PR #691, opened 2023-12-20
refactor: Make the `coordinates()` parameters nullable

sschuberth:
Allow to set only specific (non-null) arguments.


gabrielittner (approved):
Thanks

Issue #692, opened 2023-12-21
Ideas for improving the release process tasks


sschuberth:
Currently, it seems a bit awkward to me that the tasks to publish to Maven Central in any case already close the staging repository, but the (supposed to be?) complementing task is called "*closeAnd*ReleaseRepository". The obvious first question from a new user probably is: Why would I want to close my staging repository again, and why is there no "releaseRepository" only task?

Also, controlling the auto-release feature of the publish tasks via a property named `SONATYPE_AUTOMATIC_RELEASE` (which, due to its casing, looks more like an environment variable, BTW) is not very transparent, IMO.

To make this more consistent, I propose to drop the `closeAndReleaseRepository` task and instead teach the publish task to take task options named `--publish`, `--close`, and `--release` (maybe `--close` could be omitted if `--publish` is supposed to always close). That way, these options would show up with `./gradlew help --task publishTask`.

gabrielittner:
I definitely agree that this is an area that can be further improved.

> Currently, it seems a bit awkward to me that the tasks to publish to Maven Central in any case already close the staging repository, but the (supposed to be?) complementing task is called "closeAndReleaseRepository". The obvious first question from a new user probably is: Why would I want to close my staging repository again, and why is there no "releaseRepository" only task?

This is mostly a historical leftover from before we had the automatic closing at the end of each build that runs `publish`. So like you said in practice the task will actually skip the closing part and just release the repository. So renaming the task to `releaseRepository` definitely makes sense (or adding it as a second task and deprecating the old one).

> Also, controlling the auto-release feature of the publish tasks via a property named SONATYPE_AUTOMATIC_RELEASE (which, due to its casing, looks more like an environment variable, BTW) is not very transparent, IMO.

The other way to control it is by passing `true` as second parameter to `publishToMavenCentral` to permanently turn it on. The casing is for consistency with other properties and comes from the original 10+ year old script that this plugin is based on. We can also start introducing aliases that use camel case.

> To make this more consistent, I propose to drop the closeAndReleaseRepository task and instead teach the publish task to take task options named --publish, --close, and --release (maybe --close could be omitted if --publish is supposed to always close). That way, these options would show up with ./gradlew help --task publishTask.

Unfortunately we can't teach the publish task anything since it's coming from Gradle and as far as I know you can't dynamically add options to an existing task class. Not using the default tasks would mean duplicating all the uploading logic that Gradle comes with. 

[...]

PR #698, opened 2024-01-01
always close, release and drop from build service

gabrielittner:
Partially addresses #692.

Building on top of #695 this also runs the close, release and drop task actions always from the build service. Unlike the create actions these will not be run from the task. Instead the task only tells the build service to run the action and the service will run it at the end of the build. This allows doing something like `./gradlew publish closeAndReleaseRepository` (for when you don't want to permantently enable automatic releases) instead of having it as 2 separate Gradle invocations where the second needs to search for the repository again.

### DIFF
--- a/docs/central.md
+++ b/docs/central.md
@@ -279,16 +279,22 @@ The publishing process for Maven Central consists of several steps
 4. The staging repository is released
 5. All artifacts in the released repository will be synchronized to maven central
 
-By running the following Gradle task the plugin will take care of steps 1 to 3 automatically:
+The plugin will always do step 1 to 3 and when automatic releases are enabled also take
+care of step 4.
+
+Regardless of whether it is done automatically or manually after the staging repository is released
+the artifacts will be synced to Maven Central. This process takes 10-30 minutes and when it is completed
+the artifacts are available for download.
+
+### Automatic release
+
+Run the following to let the plugin handle all steps automatically:
 
 ```
-./gradlew publishAllPublicationsToMavenCentralRepository --no-configuration-cache
+./gradlew publishAllPublicationsToMavenCentralRepository closeAndReleaseRepository --no-configuration-cache
 ```
 
-The releasing step can be done manually by going to oss.sonatype.org (or s01.oss.sonatype.org) and
-clicking the button in the web UI or by executing `./gradlew closeAndReleaseRepository`.
-
-It is also possible to configure the plugin to perform this step automatically together with the first 3
+It is also possible to permanently enable the automatic releases without having to specify `closeAndReleaseRepository`
 by adding an extra parameter in the DSL or setting a Gradle property
 
 === "build.gradle"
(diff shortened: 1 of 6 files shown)


docs/central.md
```diff
@@ -279,16 +279,22 @@ The publishing process for Maven Central consists of several steps
...
+the artifacts will be synced to Maven Central. This process takes 10-30 minutes and when it is completed
+the artifacts are available for download.
+
+### Automatic release
+
+Run the following to let the plugin handle all steps automatically:
 
 ```
-./gradlew publishAllPublicationsToMavenCentralRepository --no-configuration-cache
+./gradlew publishAllPublicationsToMavenCentralRepository closeAndReleaseRepository --no-configuration-cache
```
gabrielittner, line 294: I'll do a follow up where I add `releaseRepository` as an alias for this task and a new `publishAndReleaseToMavenCentral` that depends on both `publishAllPublicationsToMavenCentralRepository` and `releaseRepository`. That way this becomes a lot shorter.

vanniktech (approved):

docs/central.md
```diff
@@ -279,16 +279,22 @@ The publishing process for Maven Central consists of several steps
 4. The staging repository is released
 5. All artifacts in the released repository will be synchronized to maven central
 
-By running the following Gradle task the plugin will take care of steps 1 to 3 automatically:
+The plugin will always do step 1 to 3 and when automatic releases are enabled also take
+care of step 4.
```
sschuberth, line 283: I propose instead:

> The plugin will always do steps 1 to 3. Step 4 is only taken care of if automatic releases are enabled.

docs/central.md
```diff
@@ -279,16 +279,22 @@ The publishing process for Maven Central consists of several steps
...
+
+Regardless of whether it is done automatically or manually after the staging repository is released
+the artifacts will be synced to Maven Central. This process takes 10-30 minutes and when it is completed
+the artifacts are available for download.
```
sschuberth, line 287: I propose instead:

> After the staging repository has been released, either manually or automatically, the artifacts will be synced to Maven Central. This process usually takes around 10-30 minutes and only when it completes the artifacts are available for download.

docs/central.md
```diff
@@ -279,16 +279,22 @@ The publishing process for Maven Central consists of several steps
...
+
+### Automatic release
+
+Run the following to let the plugin handle all steps automatically:
```
sschuberth, line 291: "... following *tasks* ..."

docs/central.md
```diff
@@ -321,11 +327,15 @@ by adding an extra parameter in the DSL or setting a Gradle property
...
 
-Regardless of whether it is done automatically or manually after the staging repository is released the artifacts
-will be synced to Maven Central. This process takes 10-30 minutes and when it is completed
-the artifacts are available for download.
+### Manual release
+
+The release (step 4) can be done manually by running the following command, so that the plugin will
+only do step 1 to 3:
+```
+./gradlew publishAllPublicationsToMavenCentralRepository --no-configuration-cache
```
sschuberth, line 335: Why is `--no-configuration-cache` required? Wouldn't it work without it? If so, why?

docs/central.md
```diff
@@ -321,11 +327,15 @@ by adding an extra parameter in the DSL or setting a Gradle property
...
+```
 
-## Timeouts
+### Timeouts
```
sschuberth, line 338: Intended change?

docs/central.md
gabrielittner: Updated 👍

docs/central.md
gabrielittner: Updated 👍

docs/central.md
gabrielittner: I've added a note to the docs about it. We're currrently being blocked by Gradle itself from supporting configuration cache for release uploads.

docs/central.md
gabrielittner: Yes, these timeouts only affect publishing releases so it makes sense to move it under that.

Issue #722, opened 2024-02-11
Support Sonatype Portal Publisher API


eygraber:
Looks like [Sonatype has a new API in preview for publishing](https://central.sonatype.org/publish/publish-portal-api/).

sschuberth:
Also see https://github.com/GradleUp/nmcp.

gabrielittner:
I've took a look at it last weekend and I'm planning to have an initial implementation done either this week or next.

gabrielittner:
Took a while but #733 addresses this.

If you want to already try it out:
- change version of the plugin to `central-portal-SNAPSHOT`
- use `SONATYPE_HOST=CENTRAL_PORTAL` or `publishToMavenCentral(SonatypeHost.CENTRAL_PORTAL, ...)`
- set the user name and password from your central portal user token like described in the existing docs

Notes:
- publishing snapshots is not supported for the central portal
- the build will currently not fail if the validations on the portal side fail after the upload succeeded

PR #724, opened 2024-02-20
First draft of Sonatype Central Portal API client

kellyrob99:
related to https://github.com/vanniktech/gradle-maven-publish-plugin/issues/722

@gabrielittner I hacked this together as part of some learning activities around Kotlin+retrofit, so I can't vouch for whether this is the best way to implement these features, but if it's at all valuable in your exploration of #722 please feel free to do with it what you will :)


gabrielittner:
I didn't do anything with the API yet, because I have been focusing on the Gradle part, so this is very useful. Could you move it to a separate module? We could call it central-portal.

ScottPierce:
I'd be down to test this on https://github.com/ScottPierce/kotlin-env-var if there is a snapshot available.

I have a new sonatype account, and I think the new sonatype release system is my only option, so I'd be down to test the new implementation.

gabrielittner:
@ScottPierce I'll share in the issue as soon as I have a working snapshot

central-portal/src/main/kotlin/com/vanniktech/maven/publish/portal/SonatypeCentralPortal.kt
```diff
@@ -97,15 +96,6 @@ class SonatypeCentralPortal(
...
   }
 
-  private fun browseDeployments(input: FileRequest): List<RepositoryActivity>? {
-    val listResponse = service.browseDeployments(input).execute()
-    if (listResponse.isSuccessful) {
-      return listResponse.body()
-    } else {
-      throw IOException("Failed to browse deployments for: $input msg: ${listResponse.errorBody()?.string()}")
-    }
-  }
```
gabrielittner: Removed this one since it made the build fail and I couldn't find it here https://central.sonatype.org/publish/publish-portal-api/#publish-or-drop-the-deployment

gabrielittner (approved):
Thanks for this

kellyrob99:
> @ScottPierce I'll share in the issue as soon as I have a working snapshot

Entirely my pleasure!

hfhbd:
Where did you find the api of the Central portal? The OpenAPI schema is a pain, many definitions are missing or only available as (wrong) example: https://central.sonatype.com/api-doc

PR #733, opened 2024-03-10
support publishing through central portal

### LINKED ISSUE #720: `Cannot get stagingProfiles for account` or `Received status code 401 from server: Unauthorized`
I'm trying to publish my lib to MavenCentral using the newest version of this plugin.

My configuration is relatively standard.
```
mavenPublishing {
//    publishToMavenCentral(SonatypeHost.DEFAULT)
// trying both
    publishToMavenCentral(SonatypeHost.S01)

    signAllPublications()

    coordinates("io.github.admund", "marketspecificui", "0.1.0-SNAPSHOT")
// trying both
//    coordinates("io.github.admund", "marketspecificui", "0.1.0")

    pom {
        name.set("MarketSpecificUI")
        description.set("The library is for helping manage market-specific resources for Android apps.")
        inceptionYear.set("2024")
        url.set("https://github.com/admund/MarketSpecificUI")
        licenses {
            license {
                name.set("The Apache License, Version 2.0")
                url.set("http://www.apache.org/licenses/LICENSE-2.0.txt")
                distribution.set("http://www.apache.org/licenses/LICENSE-2.0.txt")
            }
        }
        developers {
            developer {
                id.set("admund")
                name.set("Adrian Gardian")
                url.set("https://github.com/admund")
            }
        }
        scm {
            url.set("https://github.com/admund/MarketSpecificUI")
            connection.set("scm:git:git://github.com/admund/MarketSpecificUI.git")
            developerConnection.set("scm:git:ssh://git@github.com/admund/MarketSpecificUI.git")
        }
    }
}
```

When I'm trying to publish a version with `-SNAPSHOT`, I'm getting:
```
Could not PUT 'https://s01.oss.sonatype.org/content/repositories/snapshots/io/github/admund/marketspecificui/0.1.0-SNAPSHOT/maven-metadata.xml'. Received status code 401 from server: Unauthorized
```
But without `-SNAPSHOT`, I'm getting:
```
Cannot get stagingProfiles for account <username>: 
```
What are those `stagingProfiles`?

I was trying both `SonatypeHost.DEFAULT` and `SonatypeHost.S01`, same errors.

I created an account here https://central.sonatype.com/, and I generated a username and password, which I put in `~/.gradle/gradle.properties`.

What should I do else? I'm wondering if this is a plugin issue or maybe a MavenCentral issue?

### LINKED ISSUE #722: Support Sonatype Portal Publisher API

gabrielittner:
This is a first working implementation. There are some parts that can be improved in the future but for now I wanted to keep the code shared as much as possible. 

General process:
1) `createStagingRepsitory` creates a uuid inside the build service
2) All modules publish to `<root>/build/publish/staging>/<uuid>`
3) The build service will create a zip out of that at the end of the build
4) The build service makes the upload API call (this is now just one call in which we directly say whether we want to automatically release or not)

Also supported
- using `releaseRepository` and `dropRepository` when specifying a depoyment id through `--repository`

Not supported:
- snapshots (generally not supported by central portal)
- using `releaseRepository` and `dropRepository` without specifying a depoyment id through `--repository`

Future improvements:
- Wait for validations to finish before stopping the build (maybe also wait for publishing, but it seems slow from my test)
- Use Gradle mechanisms to publish each module into it's own build folder and make the output consumable. Then have a propert zip and upload task in the root project that consumes these.
- Add versions of `releaseRepository` and `dropRepository` that use `deployment` in their name and `deployment-id` for the parameter? Would be more aligned with the new naming.

Closes #720, closes #722

### DIFF
--- a/plugin/src/main/kotlin/com/vanniktech/maven/publish/SonatypeHost.kt
+++ b/plugin/src/main/kotlin/com/vanniktech/maven/publish/SonatypeHost.kt
@@ -8,11 +8,18 @@ import java.io.Serializable
  *
  * https://central.sonatype.org/articles/2021/Feb/23/new-users-on-s01osssonatypeorg/
  */
-data class SonatypeHost(
+data class SonatypeHost internal constructor(
   internal val rootUrl: String,
+  internal val centralPortal: Boolean,
 ) : Serializable {
+  constructor(rootUrl: String) : this(rootUrl, centralPortal = false)
+
   internal fun apiBaseUrl(): String {
-    return "$rootUrl/service/local/"
+    return if (centralPortal) {
+      "$rootUrl/api/v1/"
+    } else {
+      "$rootUrl/service/local/"
+    }
   }
 
   companion object {
@@ -20,13 +27,17 @@ data class SonatypeHost(
     fun valueOf(sonatypeHost: String): SonatypeHost = when (sonatypeHost) {
       "DEFAULT" -> DEFAULT
       "S01" -> S01
+      "CENTRAL_PORTAL" -> CENTRAL_PORTAL
       else -> throw IllegalArgumentException("No SonatypeHost constant $sonatypeHost")
     }
 
     @JvmField
-    val DEFAULT = SonatypeHost("https://oss.sonatype.org")
+    val DEFAULT = SonatypeHost("https://oss.sonatype.org", centralPortal = false)
 
     @JvmField
-    val S01 = SonatypeHost("https://s01.oss.sonatype.org")
+    val S01 = SonatypeHost("https://s01.oss.sonatype.org", centralPortal = false)
+
+    @JvmField
+    val CENTRAL_PORTAL = SonatypeHost("https://central.sonatype.com", centralPortal = true)
   }
 }
--- a/plugin/src/main/kotlin/com/vanniktech/maven/publish/SonatypeHost.kt
+++ b/plugin/src/main/kotlin/com/vanniktech/maven/publish/SonatypeHost.kt
@@ -10,12 +10,12 @@ import java.io.Serializable
  */
 data class SonatypeHost internal constructor(
   internal val rootUrl: String,
-  internal val centralPortal: Boolean,
+  internal val isCentralPortal: Boolean,
 ) : Serializable {
-  constructor(rootUrl: String) : this(rootUrl, centralPortal = false)
+  constructor(rootUrl: String) : this(rootUrl, isCentralPortal = false)
 
   internal fun apiBaseUrl(): String {
-    return if (centralPortal) {
+    return if (isCentralPortal) {
       "$rootUrl/api/v1/"
     } else {
       "$rootUrl/service/local/"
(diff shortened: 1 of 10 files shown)


plugin/src/main/kotlin/com/vanniktech/maven/publish/SonatypeHost.kt
```diff
@@ -8,25 +8,36 @@ import java.io.Serializable
  *
  * https://central.sonatype.org/articles/2021/Feb/23/new-users-on-s01osssonatypeorg/
  */
-data class SonatypeHost(
+data class SonatypeHost internal constructor(
   internal val rootUrl: String,
+  internal val centralPortal: Boolean,
```
vanniktech, line 13: Maybe?

```suggestion
  internal val isCentralPortal: Boolean,
```

`if (!host.centralPortal) {` reads very weirdly.

plugin/src/main/kotlin/com/vanniktech/maven/publish/SonatypeHost.kt
```diff
@@ -8,25 +8,36 @@ import java.io.Serializable
...
 ) : Serializable {
+  constructor(rootUrl: String) : this(rootUrl, centralPortal = false)
```
vanniktech, line 15: can be internal, right?

plugin/src/main/kotlin/com/vanniktech/maven/publish/SonatypeHost.kt
```diff
@@ -8,25 +8,36 @@ import java.io.Serializable
...
     @JvmField
-    val DEFAULT = SonatypeHost("https://oss.sonatype.org")
+    val DEFAULT = SonatypeHost("https://oss.sonatype.org", centralPortal = false)
+
+    @JvmField
+    val S01 = SonatypeHost("https://s01.oss.sonatype.org", centralPortal = false)
 
     @JvmField
-    val S01 = SonatypeHost("https://s01.oss.sonatype.org")
+    val CENTRAL_PORTAL = SonatypeHost("https://central.sonatype.com", centralPortal = true)
```
vanniktech, line 41: Hmm or should we just make it a `sealed interface` with 3 data object implementations?

plugin/src/main/kotlin/com/vanniktech/maven/publish/sonatype/SonatypeRepositoryBuildService.kt
```diff
@@ -163,14 +219,75 @@ internal abstract class SonatypeRepositoryBuildService :
...
+    if (closeActions.isNotEmpty()) {
+      if (uploadId != null) {
+        val deploymentName = "${parameters.groupId.get()}-$uploadId"
+        val publishingType = if (actions.contains(EndOfBuildAction.ReleaseAfterClose)) {
+          "AUTOMATIC"
+        } else {
+          "USER_MANAGED"
+        }
+
+        val directory = File(publishingUrl(false).substringAfter("://"))
```
vanniktech, line 249: Won't this fail on Windows?

plugin/src/main/kotlin/com/vanniktech/maven/publish/SonatypeHost.kt
gabrielittner: f842404

plugin/src/main/kotlin/com/vanniktech/maven/publish/SonatypeHost.kt
gabrielittner: No, it's public to support custom Sonatype hosts like https://google.oss.sonatype.org/

plugin/src/main/kotlin/com/vanniktech/maven/publish/SonatypeHost.kt
gabrielittner: Same as above

plugin/src/main/kotlin/com/vanniktech/maven/publish/sonatype/SonatypeRepositoryBuildService.kt
gabrielittner: You mean because of the `substringAfter("://")`? I think it should be fine since it's just removing the prefix that we add ourselves inside `publishingUrl`

vanniktech (approved):

PR #750, opened 2024-03-28
Gradle 8.8 support

gabrielittner:
Just opening this already so that I don't forget about it. Needs to wait for rc1 to be available before getting merged because the nightly version is not considered smaller than a rc version so the check wouldn't actually work.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,9 +10,9 @@
 - Kotlin Gradle Plugin 1.9.20
 
 #### Compatibility tested up to
-- JDK 22
+- JDK 21
 - Gradle 8.7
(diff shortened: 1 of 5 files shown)


vanniktech (approved):

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
 #### Compatibility tested up to
 - JDK 21
 - Gradle 8.7
+- Gradle 8.8-rc-1g
```
vanniktech, line 15:
```suggestion
- Gradle 8.8-rc-1
```

vanniktech (approved):

PR #784, opened 2024-05-22
Add status code to error message

ZacSweers:
Not all error responses have error bodies, so currently this error is opaque if there isn't one. For example - if it's a 500 error, then the user would know there's nothing they've done wrong and it's just an issue on sonatype


gabrielittner (approved):
👍

PR #788, opened 2024-05-25
fix issue with javadocJar tasks

gabrielittner:
This creates a separate javadocJar task per publication. This avoids an issue where Gradle gets confused and thinks the publish task, metadata task or sign task of one publication has an implicit dependency on the task. While this means there is a bit of additional work, zipping should generally be fast and the javadoc (or dokka) generation itself is still just run once.

### DIFF
--- a/plugin/src/main/kotlin/com/vanniktech/maven/publish/MavenPublishBaseExtension.kt
+++ b/plugin/src/main/kotlin/com/vanniktech/maven/publish/MavenPublishBaseExtension.kt
@@ -153,11 +153,6 @@ abstract class MavenPublishBaseExtension(
       project.gradleSigning.sign(publication)
     }
 
-    // TODO: https://youtrack.jetbrains.com/issue/KT-46466 https://github.com/gradle/gradle/issues/26091
-    project.tasks.withType(AbstractPublishToMaven::class.java).configureEach { publishTask ->
-      publishTask.dependsOn(project.tasks.withType(Sign::class.java))
-    }
-
     // TODO: https://youtrack.jetbrains.com/issue/KT-61313/ https://github.com/gradle/gradle/issues/26132
     project.plugins.withId("org.jetbrains.kotlin.multiplatform") {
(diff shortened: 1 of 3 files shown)


plugin/src/main/kotlin/com/vanniktech/maven/publish/MavenPublishBaseExtension.kt
```diff
@@ -153,11 +152,6 @@ abstract class MavenPublishBaseExtension(
       project.gradleSigning.sign(publication)
     }
 
-    // TODO: https://youtrack.jetbrains.com/issue/KT-46466 https://github.com/gradle/gradle/issues/26091
-    project.tasks.withType(AbstractPublishToMaven::class.java).configureEach { publishTask ->
-      publishTask.dependsOn(project.tasks.withType(Sign::class.java))
-    }
```
gabrielittner, line 159: This was just fixing the issue for the signing task but not for the publish and metadata tasks.

plugin/src/main/kotlin/com/vanniktech/maven/publish/tasks/JavadocJar.kt
```diff
@@ -14,33 +14,40 @@ open class JavadocJar : Jar() {
...
+        is JavadocJarOption.Dokka -> dokkaJavadocJar(prefix, javadocJar.taskName)
       }
     }
 
-    private fun Project.emptyJavadocJar(): TaskProvider<*> = tasks.register("emptyJavadocJar", JavadocJar::class.java)
+    private fun Project.emptyJavadocJar(prefix: String): TaskProvider<*> = tasks.register(
+      "${prefix}EmptyJavadocJar",
+      JavadocJar::class.java,
+    ) {
+      it.archiveBaseName.set("$prefix-javadoc")
```
gabrielittner, line 30: Need to set this so that the output has different names per publication, this is just the local file name and doesn't influence how it is published.

vanniktech (approved):

PR #823, opened 2024-08-11
Add note about Central Portal publishing vs OSSRH to docs

hbmartin:


gabrielittner:
Thank you. https://github.com/vanniktech/gradle-maven-publish-plugin/pull/839 also covers user tokens so I'm closing thos one

PR #839, opened 2024-09-19
Improve secrets documentation

dlew:
I made two changes here:

- Reduced some of the awkward phrasing in the explanatory secrets paragraph.

- Added a note about Sonatype user tokens.

### DIFF
--- a/docs/central.md
+++ b/docs/central.md
@@ -212,10 +212,9 @@ This configuration also determines the coordinates (`group:artifactId:version`) 
 
 ## Secrets
 
-For the publishing to work the credentials for Sonatype OSS as well as for the
-GPG key that is used for signing need to provided. To keep them out of version
-control it is recommended to either put this into the `gradle.properties` file
-user home or to use environment variables for publishing from CI servers.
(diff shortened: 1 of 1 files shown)


docs/central.md
```diff
@@ -242,6 +241,9 @@ user home or to use environment variables for publishing from CI servers.
     ORG_GRADLE_PROJECT_signingInMemoryKeyPassword=some_password
     ```
 
+Note that the username/password here is *not* the same one you use to login; Sonatype publishing
+requires a username/password that was [generated via user tokens](https://central.sonatype.org/publish/generate-portal-token/).
```
gabrielittner, line 245: Just a small change to clarify that Central Portal user tokens are separate from user tokens for the old sites:
```suggestion
requires a username/password that was generated via user tokens. The user token needs to be obtained on [Sonatype OSS](https://central.sonatype.org/publish/generate-token/) or the [Central Portal](https://central.sonatype.org/publish/generate-portal-token/) depending on where you publish.
```

gabrielittner (approved):
Thanks a lot for the addition and improvement.

PR #849, opened 2024-10-03
add support for Dokka 2

gabrielittner:
https://kotlinlang.org/docs/dokka-migration.html#select-documentation-output-format

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,6 @@
 
 - Add support for Dokka 2.0.0-Beta
     - Supports `org.jetbrains.dokka.experimental.gradle.pluginMode=V2Enabled`
-    - When `pluginMode` is set to `V2EnabledWithHelpers` the old v1 tasks are used
     - Supports both `org.jetbrains.dokka` and `org.jetbrains.dokka-javadoc`
     - If both are applied the javadoc output is published
     - Removed support for the old `org.jetbrains.dokka-android` plugin
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## 0.30.0 **UNRELEASED**
 
+- Add support for Dokka 2.0.0-Beta
+    - Supports `org.jetbrains.dokka.experimental.gradle.pluginMode=V2Enabled`
+    - When `pluginMode` is set to `V2EnabledWithHelpers` the old v1 tasks are used
+    - Supports both `org.jetbrains.dokka` and `org.jetbrains.dokka-javadoc`
+    - If both are applied the javadoc output is published
+    - Removed support for the old `org.jetbrains.dokka-android` plugin
 - Remove usages of deprecated Gradle API that is scheduled to be removed in Gradle 9.0
 - Raised minimum supported Gradle version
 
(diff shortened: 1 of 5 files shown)


plugin/src/main/kotlin/com/vanniktech/maven/publish/MavenPublishBaseExtension.kt
```diff
@@ -406,10 +406,16 @@ abstract class MavenPublishBaseExtension @Inject constructor(
...
+      JavadocJar.Dokka("dokkaGeneratePublicationJavadoc")
+    } else if (project.plugins.hasPlugin("org.jetbrains.dokka")) {
       val dokkaTask = project.provider {
         val tasks = project.tasks.withType(DokkaTask::class.java)
-        tasks.singleOrNull()?.name ?: "dokkaHtml"
+        if (tasks.size == 0) {
+          "dokkaGeneratePublicationHtml"
+        } else {
+          tasks.singleOrNull()?.name ?: "dokkaHtml"
+        }
```
gabrielittner, line 418: With `V2Enabled` there will be 0 registered `DokkaTask` instances so we can use that as an indicator. When Dokka reached 2.1.0 we can make v2 mode required and simplify all of this to just  `JavadocJar.Dokka("dokkaGeneratePublicationHtml)`, like the javadoc block above.

plugin/src/integrationTest/kotlin/com/vanniktech/maven/publish/ProjectSpecRunner.kt
```diff
@@ -41,8 +41,8 @@ fun ProjectSpec.run(fixtures: Path, temp: Path, options: TestOptions): ProjectRe
 }
 
 private fun supportsConfigCaching(plugins: List<PluginSpec>): Boolean {
-  // TODO https://github.com/Kotlin/dokka/issues/2231
-  return !plugins.any { it.id == dokkaPlugin.id }
+  // TODO can always return true when dropping support for dokka 1
+  return plugins.none { it.id == dokkaPlugin.id }
```
gabrielittner, line 45: Config cache doesn't like the `provider.flatMap { tasks.named(it) }` that we're doing right now. So until we're dropping support for Dokka 1 it will stay incompatible.

vanniktech (approved):

plugin/src/integrationTest/kotlin/com/vanniktech/maven/publish/ProjectSpecRunner.kt
whyoleg: FYI: Configuration cache is supported ONLY when pluginMode is `V2Enabled/V2EnabledWithHelpers` - so in so called DGP v2

CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 ## 0.30.0 **UNRELEASED**
 
+- Add support for Dokka 2.0.0-Beta
+    - Supports `org.jetbrains.dokka.experimental.gradle.pluginMode=V2Enabled`
+    - When `pluginMode` is set to `V2EnabledWithHelpers` the old v1 tasks are used
```
whyoleg, line 7: FYI: in case of `V2EnabledWithHelpers` we do only create stub tasks which matches old types - they will fail to generate documentation.
Still, it should not really matter, as `V2EnabledWithHelpers` should only be used during migration and should not be committed or used afterwards

whyoleg (commented):
Thanks for so fast support for DGPv2!
I've just added some more context in comments :)

plugin/src/integrationTest/kotlin/com/vanniktech/maven/publish/ProjectSpecRunner.kt
gabrielittner: 👍 I've clarified/updated the comment

Issue #798, opened 2024-06-25
How to set username and password config by project local.properties or by parameters


zhlinh:
Now provided two ways to set username and password config. by ~/.gradle/gradle.properties or by system env.

But if want different project has different username and password settings, I found it so hard to do it. 
1. I want the plugin to read config in project local.properties, but it can't
2. Or I read it by myself and set it to this plugin, but I can not find how to set it by parameters
3. It's hard to change system env in the gradle scripts which I tried, but not working
```
private fun Project.setSystemEnv() {
    // set environment variable
    val envMap = mapOf(
        "ORG_GRADLE_PROJECT_mavenCentralUsername" to getLocalProperties("mavenCentralUsername", ""),
        "ORG_GRADLE_PROJECT_mavenCentralPassword" to getLocalProperties("mavenCentralPassword", ""),
        "ORG_GRADLE_PROJECT_signingInMemoryKey" to getSigningInMemoryKey(),
        "ORG_GRADLE_PROJECT_signingInMemoryKeyId" to getLocalProperties("signing.keyId", ""),
        "ORG_GRADLE_PROJECT_signingInMemoryKeyPassword" to getLocalProperties("signing.password", "")
    )
    // or set in ~/.gradle/gradle.properties, but I do not want to share with other projects
    // mavenCentralUsername
    // mavenCentralPassword
    // signing.keyId
    // signing.password
    // signing.secretKeyRingFile
    // detail in https://vanniktech.github.io/gradle-maven-publish-plugin/central/#configuring-the-pom
    tasks.withType(Exec::class.java).configureEach {
        // set env
        environment(envMap)
        envMap.forEach { (key, value) ->
            if (value.isNotEmpty()) {
                System.setProperty(key, value)
            }
        }
    }
}
```

alirezat775:
I need the same capability, we are using `secrets.properties` to provide our credentials for the project, and since the `gradle.properties` is committed to the git, we can not save the credentials in that file, and `~/.gradle/gradle.properties` is ok but in that case we need to maintain two places for credentials, and it's preferable to have all of them in the `secrets.properties`.

basically, I need to provide a new properties file to the gradle plugin and it reads the variables from that file

```
val secretProperties = Properties().apply {
    load(FileInputStream(File(rootProject.rootDir, "secrets.properties")))
}
properties = secretProperties

```

Nek-12:
[...]

PR #867, opened 2024-11-06
Align mavenCentralUsername and mavenCentralPassword configuration with other project properties

lnhrdt:
Retains provider syntax while enabling `mavenCentralUsername` and `mavenCentralPassword` to be configured through a broader range of project properties, consistent with other properties in this plugin (e.g. `SONATYPE_HOST`).

My use case involves providing values decrypted in a Gradle task after the configuration phase, which is not currently feasible with standard Gradle properties. I believe this change would also benefit other scenarios, such as [loading values from a unique properties file](https://github.com/vanniktech/gradle-maven-publish-plugin/issues/798#issuecomment-2192088884) and other CI-centric configurations.

This PR may overlap with #824, but I believe my approach adds more general utility by allowing values to be set after the configuration phase and brings consistency to the way properties are configured across the plugin.

I attempted to run the tests using `./gradlew check`, but many failed in my development environment. However, I was able to confirm that the same tests passed and failed consistently before and after my changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Change Log
 
+## [Unreleased]
+
+- Allow mavenCentralUsername and mavenCentralPassword to be set from any property type for added flexibility.
+
 ## 0.30.0 *(2024-10-13)*
 
 - Add support for Dokka 2.0.0-Beta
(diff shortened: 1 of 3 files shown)


lnhrdt:
Hey @gabrielittner I see you closed the PR. Any feedback on the contribution? Or context to share otherwise?

gabrielittner:
Hi, sorry I missed clicking comment 😅 

The reason for closing this is that `project.properties` is not compatible with isolated projects, but `gradleProperty` is. While we can't move everything to it yet we try to have as much as possible compatible (on main the base plugin is now compatible as long as you don't use `pomFromGradleProperties`). Thanks for the contribution and sorry for taking so long to reply

PR #880, opened 2024-12-13
Correct property comment syntx in documentation

ezberlin:
The double-slash comment isn't valid in the properties syntax, so i replaced it with a #.


vanniktech (approved):

gabrielittner:
Thank you

PR #906, opened 2025-02-26
Recommend using the ASCII-armored key in its entirety

JakeWharton:
The old method did not work if you have `grep` aliased to include additional output by default (such as line numbers) and escaping any aliasing works differently in across shells. Turns out, this post-processing is not needed anyway.


gabrielittner (approved):
Thank you for making the update

PR #931, opened 2025-04-21
automate test version updates

gabrielittner:
This is done by adding extra version catalogs and then configuring renovate to allow different kinds of suffixes for each of them. We then generate constants for the version catalog version and use those in tests. The tests will also filter duplicate versions so that we don't run the same version twice if there is for example no alpha version for something.

I'm using the current versions, so that renovate should update Gradle, Kotlin and AGP to the latest versions after this is merged.


vanniktech (approved):
Neat if this works

PR #933, opened 2025-04-21
set Kotlin language level to 1.8

### LINKED ISSUE #911: Version 0.31.0 not compatible with Kotlin 1.9.20 as per release notes
The 0.31.0 release notes state that the minimum supported version is 1.9.20. 

This is not true when using a convention plugin. The following error appears:

```
Class 'com.vanniktech.maven.publish.SonatypeHost' was compiled with an incompatible version of Kotlin. The actual metadata version is 2.1.0, but the compiler version 1.9.0 can read versions up to 2.0.0.
```

gabrielittner:
Closes #911

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Fix an issue where the wrong staging profile for (s01.)oss.sonatype.org was selected
   when publishing to Maven Central.
+- Fix incompatibility with Kotlin 1.9.x when used through compiled plugins.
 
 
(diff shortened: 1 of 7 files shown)


Goooler:
It might be a bit risky to raise the language level.

https://docs.gradle.org/current/userguide/compatibility.html#kotlin

gabrielittner:
The currently published one is using 2.1, so we're lowering it here. But good point we should go with 1.8.

vanniktech (approved):

PR #956, opened 2025-05-03
Enable Explicit API mode

Goooler:
https://github.com/Kotlin/KEEP/blob/master/proposals/explicit-api-mode.md


Goooler:
Refs #955.

Issue #872, opened 2024-11-20
`findProperty()` violates isolated project constraints


solrudev:
When enabling isolated projects option for Gradle, project sync failure contains the following violation:
```
Project {project_name} cannot dynamically look up a property in the parent project :
```
Stack trace points to [this line](https://github.com/vanniktech/gradle-maven-publish-plugin/blob/691e3ec7780219299b633037c02762493e43fa06/plugin/src/main/kotlin/com/vanniktech/maven/publish/ProjectExtensions.kt#L14).

Possible replacement for `findProperty()` is `providers.gradleProperty()`.

gabrielittner:
Unfortunately `gradleProperty` and `findProperty` are not the same. `gradleProperty` will only return properties from the root project, user home or cli, while `findProperty` also returns properties from the project's `gradle.properties`. See https://github.com/gradle/gradle/issues/23572 which is effectively a blocker. We rely on being able to read project properties for configuring module specific things like the name or artifact id. 

That said some usages can probably be replaced already, so that projects that use the [base plugin](https://vanniktech.github.io/gradle-maven-publish-plugin/base/) and don't call `pomFromGradleProperties` are compatible with isolated projects.

Goooler:
Looks this has been fixed by #1181.

PR #975, opened 2025-05-17
Improve Isolated Projects compatibility

solrudev:
Make projects using base plugin and not using `gradle.properties` for POM configuration compatible with Gradle Isolated Projects (related to #872).


solrudev:
@Goooler Plugin docs suggest using only environment variables or `gradle.properties` in user's home dir for setting the affected properties, so I don't see this as blocked. It will break setups with properties declared in repo itself, but this was never recommended.

Goooler:
The main issue of `providers.gradleProperty` is that it can only read the properties under the current project but the root projects.

solrudev:
@Goooler On the contrary, `providers.gradleProperty` reads from the root project's or user's home `gradle.properties`. It doesn't read them from the current project.

Goooler:
Ooops, I misremembered that...

gabrielittner (approved):

PR #987, opened 2025-06-10
correct signing key information for maven central

brainbaking:
Hi,

The documentation is wrong with regards to signing information to maven central: 

1. You need the shorthand keyid
2. You need the ascii pass in a squashed inline form, devoid of header/footer and newlines

This had me running around in circles for a good few hours. Hopefully someone will not do that after this PR!

Thanks,
Wouter

### DIFF
--- a/docs/central.md
+++ b/docs/central.md
@@ -240,19 +240,21 @@ home `gradle.properties` file or to use environment variables (when publishing f
     ORG_GRADLE_PROJECT_mavenCentralPassword=the_password
 
     # see below for how to obtain this
-    ORG_GRADLE_PROJECT_signingInMemoryKey=exported_ascii_armored_key
-    # Optional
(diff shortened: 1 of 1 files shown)


asarkar:
I don't think the changes made in this PR are correct. I'm not using the key id, as it's optional, but I'm using the multiline value with the headers as a GitHub Secret, and it works fine in a GitHub action.

Brutus5000:
Failing with similar problems I had to remove the keyid as well to make it work (now stuck on other problems). so there is something wrong here with the key id at least.

grimsi:
The key ID has to be exactly the last 8 characters of the full key ID.
I only found this out in the Kotlin documentation which uses this plugin to publish to Maven Central: https://www.jetbrains.com/help/kotlin-multiplatform-dev/multiplatform-publish-libraries.html#add-secrets-to-github

The official docs for this plugin only have the placeholder "12345678" which isn't really obvious IMO.

Issue #978, opened 2025-05-21
Subproject doesn't use the correct version when checking for snapshot


depryf:
I have a Gradle project with multiple subprojects that need to use different versions. 

I recently updated my project to use this plugin and tried to release a snapshot on the Maven Central Portal for one of the subprojects and ran into an issue because the plugin tried to make a non-snapshot release and failed.

I then reproduced the issue on a smaller test project.

Here is the structure of the project:

```
test
 - project-a
 - project-b
```

And here is my Gradle build file:

```gradle
import com.vanniktech.maven.publish.*

plugins {
    id 'com.vanniktech.maven.publish' version '0.32.0' apply false
}

subprojects {

    apply plugin: 'java-library'
    apply plugin: 'com.vanniktech.maven.publish'

    group = 'test'
    version = project.name == 'project-a' ? '1.0' : '2.0-SNAPSHOT'

    mavenPublishing {
        configure(new JavaLibrary(new JavadocJar.Javadoc(), true))

        publishToMavenCentral(SonatypeHost.CENTRAL_PORTAL, true)
    }
}
```

So the first subproject (project-a) has a final version while the second one (project-b) is a snapshot.  

I then ran the following command to make a snapshot release of project-b:

```
> test\project-b>..\gradlew publishAllPublicationsToMavenCentralRepository
```

But I got the following error:

```
Execution failed for task ':project-b:publishMavenPublicationToMavenCentralRepository'.
> Failed to publish publication 'maven' to repository 'mavenCentral'
   > Authentication scheme 'all'(Authentication) is not supported by protocol 'file'
```

I ran the same command with --info and saw this: 

```
Publishing to repository 'mavenCentral' (file:/C:/dev/projects/test/build/publish/staging/f64bfb62-17a9-4329-85ad-7cc1b7b6dec2)
```

Which made no sense to me because snapshots are not supposed to be published locally that way.

I tried to figure out how the plugin was evaluating the version by replacing this line in the build file:

```gradle
version = project.name == 'project-a' ? '1.0' : '2.0-SNAPSHOT'
```

By this one:

```gradle
version = "${-> println('>>> getting version for subproject: ' + project.name); return project.name == 'project-a' ? '1.0' : '2.0-SNAPSHOT'}"
```

Looking at the logs, there were many entries that looked like this:

```
>>> getting version for subproject: project-b
```

Which is expected.

But there was also one entry that looked like this:

```
[...]

Issue #728, opened 2024-03-01
Publishing parts of a project with different versions leads to wrong repo being used 


gabrielittner:
Example https://github.com/cashapp/paparazzi/pull/1317/files

The actually published modules use a non SNAPSHOT version but it tries to publish to the snapshot repo because of the SNAPSHOT version in not actually published modules

gabrielittner:
As mentioned in #978 this will be resolved after the removal of the old Sonatype hosts

PR #989, opened 2025-06-13
Set credentials for Maven repository lazily

### LINKED ISSUE #988: Credentials are set eagerly despite being designed to run lazily (for the CENTRAL_PORTAL snapshot case)
Specifically in file `MavenPublishBaseExtension.kt`

You have this code (many parts elided):

```kotlin
  internal val version: Property<String> = project.objects.property(String::class.java)
    .convention(project.provider { project.version.toString() })

  public fun publishToMavenCentral(host: SonatypeHost = SonatypeHost.DEFAULT, automaticRelease: Boolean = false) {
    val versionIsSnapshot = version.map { it.endsWith("-SNAPSHOT") }
    configureCredentials { versionIsSnapshot.get() }
  }

  private inline fun configureCredentials(crossinline versionIsSnapshot: () -> Boolean) {
    project.gradlePublishing.repositories.withType(MavenArtifactRepository::class.java) { repo ->
      if (repo.name == "mavenCentral" && (!sonatypeHost.get().isCentralPortal || versionIsSnapshot())) {
        repo.credentials(PasswordCredentials::class.java)
      }
    }
  }
```

and using providers clearly demonstrates an effort to use lazy APIs. However, configureCredentials is getting called eagerly and executing the providers immediately.

I refactored out a convention plugin in my own project, and this helper plugin gets applied before my version is set...

* gradle/build-logic/src/main/kotlin/build-support.kts # common plugin
* project1/build.gradle.kts # applies build-support
* project2/build.gradle.kts # applies build-support

Both of these projects have a plugins block followed by a build script that sets the version.

```kotlin
// e.g. project1/build.gradle.kts
plugins {
    id("build-support")
}

group = "com.example"
version = libs.versions.project1.get()
```

```kotlin
// gradle/build-logic/.../build-support.kts
mavenPublishing {
    // "version" not set yet by calling build script
    publishToMavenCentral(SonatypeHost.CENTRAL_PORTAL, automaticRelease = true)
    signAllPublications()
}
```

Unfortunately, when `configureCredentials` gets called by `publishToMavenCentral`, it eagerly evaluates all those lazy provider values, which returns a version value of "unspecified" and not "1.2.3-SNAPSHOT" or whatever.

## Possible fix?

I feel like using `afterEvaluate` is a hack, but I'm pretty sure this would work:

```kotlin
  private inline fun configureCredentials(crossinline versionIsSnapshot: () -> Boolean) {
    project.afterEvaluate {
       project.gradlePublishing.repositories.withType(MavenArtifactRepository::class.java) { repo ->
         if (repo.name == "mavenCentral" && (!sonatypeHost.get().isCentralPortal || versionIsSnapshot())) {
           repo.credentials(PasswordCredentials::class.java)
         }
       }
     }
   }
```

solrudev:
Fixes #988.

### DIFF
--- a/plugin/src/main/kotlin/com/vanniktech/maven/publish/MavenPublishBaseExtension.kt
+++ b/plugin/src/main/kotlin/com/vanniktech/maven/publish/MavenPublishBaseExtension.kt
@@ -86,7 +86,13 @@ public abstract class MavenPublishBaseExtension @Inject constructor(
       repo.setUrl(buildService.map { it.publishingUrl() })
     }
 
-    configureCredentials { versionIsSnapshot.get() }
+    project.afterEvaluate { project ->
+      project.gradlePublishing.repositories.withType(MavenArtifactRepository::class.java).configureEach { repo ->
+        if (repo.name == "mavenCentral" && (!sonatypeHost.get().isCentralPortal || versionIsSnapshot.get())) {
+          repo.credentials(PasswordCredentials::class.java)
+        }
+      }
+    }
 
     val createRepository = project.tasks.registerCreateRepository(buildService, groupId, artifactId, version)
 
@@ -237,8 +243,6 @@ public abstract class MavenPublishBaseExtension @Inject constructor(
     project.mavenPublications {
       it.version = this.version.get()
     }
-
-    configureCredentials { version.endsWith("-SNAPSHOT") }
   }
 
   /**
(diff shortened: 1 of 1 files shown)


gabrielittner:
I think it would be better to just go with wrapping it in `afterEvaluate`. 
- when OSSRH support is removed we can have `setUrl` also there and resolve the config cache incompatibility (since Gradle doesn't seem to move on fixing/adding provider support for that)
- I'm not sure if with this change the provider might be evaluated too late. Since it's only accessed in a build service it might only happen during execution at least with config cache disabled (not 100% sure on this)

gabrielittner:
We can still remove it from the `version` method though since it's then lazy anyways.

solrudev:
> We can still remove it from the `version` method though since it's then lazy anyways.

Do you mean we can remove it after wrapping `configureCredentials` in `afterEvaluate`?

gabrielittner:
Yes, or just have `afterEvaluate { project.gradlePublishing.repositories.withType(MavenArtifactRepository::class.java).configureEach { repo -> ...` directly inside `publishToMavenCentral`

solrudev:
You are right. Putting this inside of `versionIsSnapshot` provider is problematic for multi-project setups, as only one project's provider will be evaluated when registering the build service, so other projects will not have the credentials configured. I will re-do this using `afterEvaluate`.

gabrielittner (approved):
Thank you

solrudev:
`afterEvaluate` workaround will do. But to handle this properly it's worth trying to tackle #728, also see #978.

gabrielittner:
👍 just commented on the latter issue that supporting different versions will be possible after the shut down of the old Sonatype hosts.

PR #992, opened 2025-06-18
Deprecate DEFAULT and S01

### LINKED ISSUE #954: Deprecate the usages of `DEFAULT` and `S01`
Users should migrate to the central portal in the future due to the planned sunset of OSSRH.
We can deprecate the use of `DEFAULT` and `S01` to warn users about the upcoming migration in future versions.

https://central.sonatype.org/news/20250326_ossrh_sunset/

Goooler:
Closes #954.


Goooler:
Probably we should remove these properties in the next release.

gabrielittner (approved):

PR #993, opened 2025-06-18
Link releases in changelog

Goooler:
Keep the file closer to https://keepachangelog.com/en/1.1.0/.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 # Change Log
 
-## 0.32.0 **UNRELEASED**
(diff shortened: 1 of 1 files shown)


gabrielittner (approved):
Thank you

Goooler:
I see https://github.com/vanniktech/gradle-maven-publish-plugin/actions/runs/15808123253/job/44555735436 didn't create the GH release automatically, may be due to this change, as 

https://github.com/vanniktech/gradle-maven-publish-plugin/blob/71e268e285f1a9f6bf52b14181f6169c39c3172d/.github/workflows/publish-release.yml#L26-L30

was empty.

We can replace this step with something like https://github.com/GradleUp/shadow/blob/2def703d1a3c505253e19be14fc53b09274e89e9/.github/workflows/release.yml#L34-L38

PR #998, opened 2025-06-21
Migrate from KAPT to KSP

solrudev:


gabrielittner:
Thanks for taking care of this, the ktlint issues would be resolved by #1000

Goooler:
I believe Moshi's reflection-based deserialization is good enough, we don't need either KAPT or KSP.

gabrielittner:
I'm fine with either. The ktlint issue is resolved on main.

PR #999, opened 2025-06-21
add support for com.android.fused-library

### LINKED ISSUE #985: Support Android Fused Library plugin
Try things like

```kt
plugins {
    alias(libs.plugins.androidFusedlibrary)
    alias(libs.plugins.mavenPublish)
}
```

There will be an exception like

```
An exception occurred applying plugin request [id: 'com.vanniktech.maven.publish', version: '0.32.0']
> Failed to apply plugin class 'org.gradle.language.base.plugins.LifecycleBasePlugin'.
   > Cannot add task 'assemble' as a task with that name already exists.
```

The Fused plugin is like a Shadow plugin for Android. 

https://developer.android.com/build/publish-library/fused-library

gabrielittner:
Closes #985 

Fairly limited support because of restrictions on the fused library part, but it's generally working.

### DIFF
--- a/docs/what.md
+++ b/docs/what.md
@@ -103,6 +103,46 @@ the specified variant instead of publishing all of them.
     }
     ```
 
+
+## Android Fused Library
+
+For projects using the `com.android.fused-library` plugin.
+
+=== "build.gradle"
+
+    ```groovy
+    import com.vanniktech.maven.publish.AndroidSingleVariantLibrary
+
+    mavenPublishing {
+      configure(new AndroidFusedLibrary())
+    }
+    ```
+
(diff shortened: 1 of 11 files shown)


docs/what.md
```diff
@@ -103,6 +103,46 @@ the specified variant instead of publishing all of them.
...
+    ```
+
+!!! warning
+
+    Support for Android fused libraries is considered experimental and might break
+    with future Android Gradle plugin updates.
+
+!!! warning
+
+    Signing is currently unsupported for Android fused libraries.
```
msfjarvis, line 138: Is there an issue tracker link that we can follow for updates on this from the AGP side?

PR #1000, opened 2025-06-21
replace ktlint plugin

gabrielittner:
Replaces running ktlint through Gradle with a Kotlin script, main motivation right now being that the ktlint plugin picks up sources generated by KSP (see #998).

Also brings in the latest ktlint version and some formatting changes related to that, some are a bit questionable but nothing too bad.


Goooler:
> ktlint plugin picks up sources generated by KSP

We can configure it to ignore sources under `projectDir/build`.

Or migrate to Spotless.

gabrielittner:
> Or migrate to Spotless.

No strong objection against sportless but last time I tried it there were some weird issues around editorconfig.

PR #1013, opened 2025-06-22
fix Github release creation

gabrielittner:
With #993 there is now a `[` between `##` and the version which is why the last release didn't get an automatic release.


Goooler:
I'm addressing another fix to #1014.

PR #1065, opened 2025-07-11
Add `--warning-mode=fail` for tests

Goooler:
---


Goooler:
Seems we can't enable this for now, as we test on various AGP / KGP and other plugins, warnings come from them.

PR #1071, opened 2025-07-13
Use per-project repositories for Central Portal publishing

### LINKED ISSUE #259: Configuration cache problems
Looks like there are some issues with using this plugin when the configuration cache is enabled. Not sure if any of this is internal to Gradle or part of the plugin, but figured I'd record it here.

[Here's a build of mine that failed because of it.](https://github.com/eygraber/kodein-di-android-components/runs/2635880947)

### LINKED ISSUE #728: Publishing parts of a project with different versions leads to wrong repo being used

### LINKED ISSUE #978: Subproject doesn't use the correct version when checking for snapshot

gabrielittner:
- Each project publishes to a local directory in its own build folder
- Each project reports that folder to the build service
- When reporting the folder it also passes the version and we ignore snapshots
- The build service collects the artifacts from all reported folders

Advantages
- Configuration cache for publishing is now fully supported because we don't rely on setUrl being lazy anymore (closes #259)
- Supports some artifacts having a final and some having a SNASPHOT and some a final version. Only the final ones will be uploaded to Central Portal (closes #728, closes #978)

Other
- Instead of having a UUID for the publishing repo dir CreateStagingRepositoryTask will now just delete the local folder to ensure its fresh every time

### DIFF
--- a/plugin/src/main/kotlin/com/vanniktech/maven/publish/sonatype/CreateSonatypeRepositoryTask.kt
+++ b/plugin/src/main/kotlin/com/vanniktech/maven/publish/sonatype/CreateSonatypeRepositoryTask.kt
@@ -35,8 +35,7 @@ internal abstract class CreateSonatypeRepositoryTask : DefaultTask() {
   fun registerProject() {
     val localRepository = localRepository.asFile.get()
 
-    // delete local repository from previous publishing attempts to ensure only current files
-    // are published
+    // delete local repository from previous publishing attempts to ensure only current files are published.
     if (localRepository.exists()) {
       localRepository.deleteRecursively()
     }
--- a/plugin/src/main/kotlin/com/vanniktech/maven/publish/sonatype/CreateSonatypeRepositoryTask.kt
+++ b/plugin/src/main/kotlin/com/vanniktech/maven/publish/sonatype/CreateSonatypeRepositoryTask.kt
@@ -1,10 +1,15 @@
 package com.vanniktech.maven.publish.sonatype
 
 import org.gradle.api.DefaultTask
+import org.gradle.api.file.Directory
+import org.gradle.api.file.DirectoryProperty
 import org.gradle.api.provider.Property
 import org.gradle.api.provider.Provider
 import org.gradle.api.tasks.Input
+import org.gradle.api.tasks.InputDirectory
 import org.gradle.api.tasks.Internal
+import org.gradle.api.tasks.PathSensitive
+import org.gradle.api.tasks.PathSensitivity
 import org.gradle.api.tasks.TaskAction
 import org.gradle.api.tasks.TaskContainer
 import org.gradle.api.tasks.TaskProvider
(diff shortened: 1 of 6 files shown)


plugin/src/main/kotlin/com/vanniktech/maven/publish/sonatype/CreateSonatypeRepositoryTask.kt
```diff
@@ -19,12 +24,24 @@ internal abstract class CreateSonatypeRepositoryTask : DefaultTask() {
...
   @get:Internal
   abstract val buildService: Property<SonatypeRepositoryBuildService>
 
   @TaskAction
   fun registerProject() {
-    buildService.get().registerProject(projectGroup.get(), artifactId.get(), version.get())
+    val localRepository = localRepository.asFile.get()
+
+    // delete local repository from previous publishing attempts to ensure only current files
+    // are published
```
Goooler, line 39:
```suggestion
    // delete local repository from previous publishing attempts to ensure only current files are published.
```

plugin/src/main/kotlin/com/vanniktech/maven/publish/sonatype/SonatypeRepositoryBuildService.kt
```diff
@@ -82,36 +66,30 @@ internal abstract class SonatypeRepositoryBuildService :
...
+    val project = MavenCentralProject(coordinates, localRepository)
+    projectsToPublish.add(project)
 
     endOfBuildActions += EndOfBuildAction.Upload
     endOfBuildActions += EndOfBuildAction.Drop(runAfterFailure = true)
   }
 
   /**
-   * Is only be allowed to be called from task actions.
+   * Is only allowed to be called from task actions.
```
Goooler, line 85: Or we can add an annotation to replace these comments.

Goooler (approved):

vanniktech (approved):

PR #1089, opened 2025-07-17
Remove unused SONATYPE_CLOSE_TIMEOUT_SECONDS and minor cleanups

Goooler:
---

### DIFF
--- a/central-portal/src/main/kotlin/com/vanniktech/maven/publish/portal/SonatypeCentralPortalOkHttpInterceptor.kt
+++ b/central-portal/src/main/kotlin/com/vanniktech/maven/publish/portal/SonatypeCentralPortalOkHttpInterceptor.kt
@@ -12,7 +12,7 @@ internal class SonatypeCentralPortalOkHttpInterceptor(
     val requestBuilder = chain
       .request()
       .newBuilder()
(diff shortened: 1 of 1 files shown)


central-portal/src/main/kotlin/com/vanniktech/maven/publish/portal/SonatypeCentralPortal.kt
```diff
@@ -12,19 +12,18 @@ import retrofit2.converter.scalars.ScalarsConverterFactory
 
 public class SonatypeCentralPortal(
   private val baseUrl: String,
-  private val usertoken: String,
+  private val userToken: String,
   userAgentName: String,
   userAgentVersion: String,
-  okhttpTimeoutSeconds: Long,
-  private val closeTimeoutSeconds: Long,
+  okhttpTimeout: Duration,
```
Goooler, line 18: This is a breaking change, but since no users are expected to rely on it, it's safe to land.

Goooler: Reverted the usage of `Duration` due to #1144.

joshfriend:
I started using this value in #1178 as the overall timeout for validating deployments

PR #1154, opened 2025-09-29
Merge central-portal project into the plugin

Goooler:
All plugin code be bundled in the plugin project, we have no need to publish `com.vanniktech:central-portal` anymore.

---


gabrielittner:
The reason for it being separate that at some point someone requested it so that they can use it outside of Gradle. While this was for nexus and I'm not sure if anyone actually uses the extra artifact, the extra maintenance of 2 modules is so low that I prefer keeping it this way.

Goooler:
I noticed `com.vanniktech:central-portal` is only depended on by the Gradle plugin, it's safe to merge them for now. We can eliminate the `build-logic` and simplify the plugin project structure in follow-ups. This is also the reason why I commented https://github.com/vanniktech/gradle-maven-publish-plugin/pull/1089#discussion_r2214671947.

https://central.sonatype.com/artifact/com.vanniktech/central-portal/versions

image

Goooler:
@gabrielittner I found the original issue is #265. It should be safe for us to land this for now.

Goooler:
We can also eliminate https://github.com/vanniktech/gradle-maven-publish-plugin/blob/bb77fa06e900ea93669ca4135a70af97aa3b3374/central-portal/build.gradle.kts#L10 after this merging.

PR #1181, opened 2025-11-05
Don't find POM properties from project hierarchy when Isolated Projects is enabled

joshfriend:
The current use of `findProperty()` sometimes causes isolated projects violations. This PR disables the hierarchal behavior and searches only the root properties via the provider api.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 - Fix misleading error message when Android library variant is not found.
 - Downgrade transitive OkHttp version.
 - Add support for validating deployments to Central Portal
+- Don't check project heirarchy for POM properties when Isolated proejcts is enabled.
 
 #### Minimum supported versions
 - JDK 11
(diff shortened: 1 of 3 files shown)


eygraber:
Is it possible to read the current project's `gradle.properties` manually when isolated projects is enabled? It would allow enabling isolated projects without having to do a whole refactor.

My use case is that I use the root `gradle.properties` for common POM values, and then each publishable project has its own `gradle.properties` with POM values specific to that artifact.

Goooler:
We can wait for https://github.com/gradle/gradle/issues/23572.

Goooler:
I did small tweaks in #1210, feel free to take a review.

eygraber:
I think there was a separate issue where Gradle said they won't be addressing this.

Goooler:
I noticed the issue 23572 was planned on Gradle 9.3, now it's moved to 9.x. I guess it will be fixed by soon.

eygraber:
I was referring to https://github.com/gradle/gradle/issues/35029#issuecomment-3321508971

> > So actually this best practice is only there to work-around https://github.com/gradle/gradle/issues/23572 that is currently scheduled to be fixed in 9.3.0, right?

> I think we know from experience that "scheduled to be fixed in 9.3.0" is no guarantee 😄. This buggy inconsistency might also be deprecated and removed in 10.0.0. We have had discussions about doing this, it just hasn't risen to the top of the heap yet.

PR #1187, opened 2025-11-06
fix test matrix

gabrielittner:
#1173 changed to a Gradle property but the CI config was passing `testConfigMethod` as System property. meaning each shard was running all 3 methods. Also changed `testJavaVersion` to a Gradle property for consistency.


Goooler:
Thanks for the fix! I missed them....

PR #1196, opened 2025-11-11
Unify min AGP and min KGP versions

Goooler:
---


gabrielittner (approved):
FYI, after 0.35.0 is released I will merge this and your JDK 17 PR. I already have a branch locally that bumps other minimum versions with some clean up.

PR #1198, opened 2025-11-11
bump minimum versions and cleanup

gabrielittner:
New minimum versions:
- Gradle 9.0.0
- Android Gradle Plugin 8.13.0
- Kotlin Gradle Plugin 2.2.0

Additionally Dokka v2 mode is now required and the Kotlin langugage level has been updated.

---

- [x] [CHANGELOG](https://github.com/vanniktech/gradle-maven-publish-plugin/blob/main/CHANGELOG.md)'s "Unreleased" section has been updated, if applicable.

PR #1209, opened 2025-11-15
Pin on Okhttp 5.1.0 and Retrofit 3.0.0

### LINKED ISSUE #1131: okhttp 5 Kotlin metadata is too new for "compatible" Gradle versions
This plugin uses okhttp 5. This is built with Kotlin 2.2.0 (with language version 2.2).

Although this plugin sets language version 1.8, the dependency on okhttp means users must have a Gradle version that embeds a Kotlin version that is compatible with metadata version 2.2 to avoid issues.

In default configurations this is not an issue. [Gradle 8.5](https://docs.gradle.org/8.5/release-notes.html#kotlin-dsl-improvements) introduced a new flag to disable skipping these checks: `org.gradle.kotlin.dsl.skipMetadataVersionCheck`. When this is set to false these metadata compatibility checks are no longer skipped as they have been in previous versions, and this plugin cannot be used in Gradle versions older than 9.0.0.

Options:
* Bump minimum Gradle version to 9.0 which embeds Kotlin 2.2
* Downgrade okhttp to 4.12
* Use a different library

Goooler:
- Closes #1131.
- Reverts part of #1144.
- Follow up #1198.

---

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -30,8 +30,6 @@ android-pluginApi = { module = "com.android.tools.build:gradle-api", version.ref
 maven-publish-plugin = "com.vanniktech:gradle-maven-publish-plugin:0.35.0"
 
 kotlin-stdlib = { module = "org.jetbrains.kotlin:kotlin-stdlib", version.ref = "kotlin" }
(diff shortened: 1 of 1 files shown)


Goooler:
Disabling updates for them is still working by

https://github.com/vanniktech/gradle-maven-publish-plugin/blob/7cb24297b2803d5b953dbb95dec30c7d87e2645b/.github/renovate.json5#L13-L19

PR #1210, opened 2025-11-15
Introduce gradlePropertyCompat

Goooler:
---


gabrielittner:
I think this defeats the purpose of the original change a bit. The properties that were just using `gradleProperty` directly are only needed on the top level so going to the compat version might introduce incompatibilties again and also some logical inconsistencies. For the POM related properties where we are using `findProperty` when isolation is enabled with the fallback we'd still have the issue for properties that are optional. So for now I would keep everything as is and hopefully Gradle will resolve this.

PR #1228, opened 2025-11-20
consistent javadoc jar option

### LINKED ISSUE #1202: No way to publish empty javadoc jar on single-variant Android project
The existing platforms are a bit restrictive. I cannot see an easy way to use `Javadoc.Empty` with an Android module, for example. Passing `javadocJar = false` will fail to publish _any_ jar, which will then fail deployment. I could attach an empty one myself, I suppose, but since the plugin knows how to do this in other contexts, I would like to leverage that.

In general, I think both an empty source jar and an empty javadoc jar should be fairly easy to configure for all module types. At this point I don't think I want to publish javadoc jars for _any_ module.

gabrielittner:
All platforms (if possible) now expose the javadoc option. Primarily exposes using `Empty` and `Dokka` for Android projects. 

Closes #1202

### DIFF
--- a/plugin/src/main/kotlin/com/vanniktech/maven/publish/Platform.kt
+++ b/plugin/src/main/kotlin/com/vanniktech/maven/publish/Platform.kt
@@ -301,7 +301,7 @@ public data class AndroidMultiVariantLibrary @JvmOverloads constructor(
 @Incubating
 public class AndroidFusedLibrary : Platform() {
   override val javadocJar: JavadocJar = JavadocJar.Empty()
-  override val sourcesJar: Boolean = true
+  override val sourcesJar: Boolean = false
 
   override fun configure(project: Project) {
     check(project.plugins.hasPlugin("com.android.fused-library")) {


Goooler (approved):

vanniktech (approved):

plugin/src/main/kotlin/com/vanniktech/maven/publish/Platform.kt
```diff
@@ -145,11 +145,21 @@ public class GradlePublishPlugin : Platform() {
  *```
  */
 public data class AndroidSingleVariantLibrary @JvmOverloads constructor(
-  val variant: String = "release",
+  override val javadocJar: JavadocJar = JavadocJar.Empty(),
   override val sourcesJar: Boolean = true,
-  val publishJavadocJar: Boolean = true,
+  val variant: String = "release",
```
Goooler: `javadocJar` should be `Javadoc` here, instead of `Empty`.

PR #1237, opened 2025-11-24
Explicitly add required dependencies into integrationTestImplementation

Goooler:
---


Goooler:
This is also useful when we are ready to re-land #1154.

PR #1246, opened 2025-11-29
support combining publishToMavenCentral with properties for automatic release and validation

gabrielittner:
Allows manually calling `publishToMavenCentral`  but still use properties for the options for `automaticRelease` and `validateDeployment`.

### DIFF
--- a/plugin/src/main/kotlin/com/vanniktech/maven/publish/Properties.kt
+++ b/plugin/src/main/kotlin/com/vanniktech/maven/publish/Properties.kt
@@ -1,0 +1,45 @@
+package com.vanniktech.maven.publish
+
+import kotlin.text.toBoolean
+import org.gradle.api.Project
+
+internal fun Project.mavenCentralPublishing(): Boolean {
+  val central = providers.gradleProperty("mavenCentralPublishing").orNull
+  if (central != null) {
+    return central.toBoolean()
+  }
+  return when (providers.gradleProperty("SONATYPE_HOST").orNull) {
+    null -> false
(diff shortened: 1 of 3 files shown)


plugin/src/main/kotlin/com/vanniktech/maven/publish/Properties.kt
```diff
@@ -0,0 +1,45 @@
...
+      """.trimIndent(),
+    )
+  }
+}
+
+internal fun Project.automaticRelease(): Boolean {
+  val automatic = providers.gradleProperty("mavenCentralAutomaticPublishing").orNull
+  if (automatic != null) {
+    return automatic.toBoolean()
+  }
```
Goooler:
```suggestion
  providers.gradleProperty("mavenCentralAutomaticPublishing").orNull?.let {
    return it.toBoolean()
  }
```

plugin/src/main/kotlin/com/vanniktech/maven/publish/Properties.kt
```diff
@@ -0,0 +1,45 @@
...
+  return providers.gradleProperty("SONATYPE_DEPLOYMENT_VALIDATION").getOrElse("true").toBoolean()
+}
+
+internal fun Project.signAllPublications(): Boolean {
+  val sign = providers.gradleProperty("signAllPublications").orNull
+  if (sign != null) {
+    return sign.toBoolean()
+  }
```
Goooler:
```suggestion
  providers.gradleProperty("signAllPublications").orNull?.let {
    return it.toBoolean()
  }
```

PR #1248, opened 2025-11-29
validate username and password being present

gabrielittner:
Should result in better error messages instead of being tied to "Failed stopped the build service" kind of errors.


plugin/src/main/kotlin/com/vanniktech/maven/publish/central/PrepareMavenCentralPublishingTask.kt
```diff
@@ -38,6 +38,21 @@ internal abstract class PrepareMavenCentralPublishingTask : DefaultTask() {
       localRepository.deleteRecursively()
     }
 
+    check(
+      buildService
+        .get()
+        .parameters.repositoryUsername.isPresent,
+    ) {
+      "Did not find mavenCentralUsername which is required for publishing to maven central"
```
Goooler:
```suggestion
      "mavenCentralUsername not found, which is required for publishing to Maven Central."
```

PR #1250, opened 2025-11-30
Support passing lazy properties to MavenPublishBaseExtension.coordinates

### LINKED ISSUE #1220: Providing version lazily in a configuration-cache compatible manner
Currently, the only way to specify the version seems to be using `MavenPublishBaseExtension.coordinates` which accepts the version only as an eager `String`.

This means that the name has to be eagerly calculated during the configuration phase. Consequently, any changes to the inputs of the version (such as a git hash) will invalidate the configuration cache.

Any chance that we could provide the version lazily, e.g. via `Provider<String>`?

Goooler:
Closes #1220.

---


gabrielittner:
This does not resolve what the issue wants to achieve because the underlying maven publication does not use properties, so it's still evaluated at configuration time.

PR #1354, opened 2026-04-02
Log deployment id

thoutbeckers:
This logs the deployment id so it can be grabbed from the log output.

I can imagine other ways to do this (task output etc) but this simple change would work for me (grabbing the id from the log and then later initiating manual actions for publishing).

It seems worth it to log this at `lifecycle` as the line `Skipping deployment validation!` below does it too.


Goooler (approved):

gabrielittner (approved):

Goooler:
@gabrielittner This needs a approval for CI running.

Issue #1161, opened 2025-10-02
Inject username & password properties


mtctx:
Hello, I am writing my own Gradle Plugin (UniPub) and I am trying to inject values for mavenCentralUsername/Password, but I cant seem to get it to work, could you help me out here?

<details>
<summary>Code Snippet</summary>

```kotlin
    override fun apply(target: Project) {
        val extension = target.extensions.create<UniPubExtension>("unipub")

        target.tasks.register<UniPubGenerateTemplateUniPubInProjectDirTask>("generateUniPubFileInProjectDir") {
            group = "UniPub"
            description = "Generate a template file for UniPub in the project directory."
        }
        target.tasks.register<UniPubGenerateTemplateUniPubInHomeDirTask>("generateUniPubFileInHomeDir") {
            group = "UniPub"
            description = "Generate a template file for UniPub at $globalPath."
        }


        val settings =
            loadAndValidateSettingsFile(extension, projectPath(target.project.projectDir.absolutePath))
        val profile = settings.profiles.find { it.name == extension.profileName }
            ?: throw GradleException(
                """
                        Profile '${extension.profileName}' not found in settings file. Please check your settings file and try again.'
                        Available profiles: ${settings.profiles.joinToString(", ") { it.name }}
                        """.trimIndent()
            )

        val usesVanniktechMavenPublish = target.plugins.hasPlugin("com.vanniktech.maven.publish")
        if (usesVanniktechMavenPublish) {
            target.setProperty("mavenCentralUsername", profile.username)
            target.setProperty("mavenCentralPassword", profile.password)
            target.logger.lifecycle("UniPub: Injected Gradle properties for Vanniktech (profile '${profile.name}')")
        }

        target.tasks.withType(PublishToMavenRepository::class.java).configureEach {
            doFirst {
                if (repository.url.scheme == "file") return@doFirst

                repository.credentials {
                    username = username?.takeIf { it.isNotBlank() && it.isNotEmpty() } ?: profile.username
                    password = password?.takeIf { it.isNotBlank() && it.isNotEmpty() } ?: profile.password
                }

                logger.lifecycle("UniPub: Injected credentials for profile '${profile.name}'")
            }
        }
    }
```
</details>
<details>
<summary>Log</summary>
23:04:02: Executing 'publish'…

> Task :checkKotlinGradlePluginConfigurationErrors SKIPPED
> Task :processResources NO-SOURCE
> Task :sourcesJar UP-TO-DATE
[...]

Issue #459, opened 2022-12-10
[Request] Add mavenCentralUsername and mavenCentralPassword as publishToMavenCentral parameters


GuilhE:
I want to move `mavenCentralUsername` and `mavenCentralPassword` from the global `~/.gradle/gradle.properties` to my project `local.properties`. I could accomplish it for the signing credentials:

```
signing {
    val signingInMemoryKeyId: String
    val signingInMemoryKey: String
    val signingInMemoryKeyPassword: String

    if (project.rootProject.file("local.properties").exists()) {
        with(Properties().apply { load(FileInputStream(File(rootProject.rootDir, "local.properties"))) }) {
            signingInMemoryKeyId = getProperty("signingInMemoryKeyId")
            signingInMemoryKey = getProperty("signingInMemoryKey")
            signingInMemoryKeyPassword = getProperty("signingInMemoryKeyPassword")
        }
    } else {
        signingInMemoryKeyId = System.getenv("ORG_GRADLE_PROJECT_signingInMemoryKeyId")
        signingInMemoryKey = System.getenv("ORG_GRADLE_PROJECT_signingInMemoryKeyId")
        signingInMemoryKeyPassword = System.getenv("ORG_GRADLE_PROJECT_signingInMemoryKeyPassword")
    }
    useInMemoryPgpKeys(signingInMemoryKeyId, signingInMemoryKey, signingInMemoryKeyPassword)
}
```
But since I cannot provide this parameters like in `useInMemoryPgpKeys()` and looking at [this](https://tomgregory.com/gradle-project-properties-best-practices/):
![imagem](https://user-images.githubusercontent.com/2677139/206879278-42ef458a-4d72-40d4-b319-30ee7ff16048.png)

It doesn't seem to be possible. Unless of course you update `publishToMavenCentral` to allow passing this values as parameters (leaving the current logic as default). Actually the same logic could be applied to `signAllPublications()` because now I'm using the `signing` extension which will override yours I believe.

gabrielittner:
We are using Gradle's credential API to configure the repository as described here https://docs.gradle.org/current/samples/sample_publishing_credentials.html. Please file a feature request with Gradle to support a project local way of providing credentials

Issue #281, opened 2021-07-20
[Feature request] An option that read maven central passord from local.properties


ichenhe:
I know it is recommended to set them in ~/.gradle/gradle.properties.

But if I have more than one maven central account, things get troublesome. So I prefer put them in `local.properties` to the main properties file.

Favorlock:
I'd like to second this request. I was doing the exact same thing before with an older version of the plugin and I have yet to figure out how to do this with the latest version. Would be great if a way to programmatically set the username/password was possible.

Favorlock:
After some digging I finally figured out how this can be done.

```groovy
ext.local = new Properties()
File file = project.rootProject.file('local.properties')
if (file.exists()) local.load(file.newDataInputStream())
    
publishing {
    repositories {
        maven {
            def releaseRepoUrl = 'https://oss.sonatype.org/service/local/staging/deploy/maven2/'
            def snapshotRepoUrl = 'https://oss.sonatype.org/content/repositories/snapshots/'
            url = version.endsWith('SNAPSHOT') ? snapshotRepoUrl : releaseRepoUrl
            credentials {
                username = local.getProperty('deployment.username')
                password = local.getProperty('deployment.password')
            }
        }
    }
}
```

gabrielittner:
I think it would be best to open a feature request with Gradle for this. We are just using their credentials API, so it would need to be something that needs to be supported on their side.

martinbonnin:
@ichenhe @Favorlock FWIW, you can also use environment variables to specify the password and username:

```
export ORG_GRADLE_PROJECT_mavenCentralUsername=$username
export ORG_GRADLE_PROJECT_mavenCentralPassword=$password
```

If you put this in a file `personnal_credential.sh` (and other credentials in e.g. `company_credential.sh`, you can source the appropriate file before releasing. Also CI usually use environment variables so this setup si closer to something you can use on Github Actions for an example.

ichenhe:
@martinbonnin Thanks. But, environment variables seem to be more suitable for CI. It is not applicable in ordinary development.

Of course, publishing to maven is not a very frequent operation, so thank you for your tip.

PR #1367, opened 2026-05-21
Allow setting Maven Central credentials via project extra properties

### LINKED ISSUE #798: How to set username and password config by project local.properties or by parameters

### LINKED ISSUE #887: Unable to specify username/password through build script
Despite setting mavenCentralUsername in project.ext, it seems gradleProperty ignores that when resolving properties. As publishToMavenCentral doesn't provide a field to set username/password, there is no way to configure it from the build script.

ingokegel:
`mavenCentralUsername` / `mavenCentralPassword` can now also be set via `extra["mavenCentralUsername"]` / `extra["mavenCentralPassword"]` on the project, in addition to the existing Gradle property sources (`gradle.properties`, `-P`, `ORG_GRADLE_PROJECT_*`). Gradle properties still take precedence, the extras lookup is only used as a fallback. This makes it possible to set credentials from a secret manager inside the build script.

Fixes #798
Fixes #887
Refs #281, #459, #1161

---

### DIFF
--- a/plugin/src/main/kotlin/com/vanniktech/maven/publish/MavenPublishBaseExtension.kt
+++ b/plugin/src/main/kotlin/com/vanniktech/maven/publish/MavenPublishBaseExtension.kt
@@ -13,6 +13,7 @@ import org.gradle.api.configuration.BuildFeatures
 import org.gradle.api.credentials.PasswordCredentials
 import org.gradle.api.plugins.ExtraPropertiesExtension
 import org.gradle.api.provider.Property
+import org.gradle.api.provider.Provider
 import org.gradle.api.publish.maven.MavenPom
 import org.gradle.api.publish.maven.MavenPublication
 import org.gradle.api.publish.maven.tasks.PublishToMavenRepository
@@ -120,8 +121,8 @@ public abstract class MavenPublishBaseExtension @Inject constructor(
     }
 
     val buildService = project.registerMavenCentralBuildService(
-      repositoryUsername = project.providers.gradleProperty("mavenCentralUsername"),
-      repositoryPassword = project.providers.gradleProperty("mavenCentralPassword"),
+      repositoryUsername = credentialProperty("mavenCentralUsername"),
+      repositoryPassword = credentialProperty("mavenCentralPassword"),
       rootBuildDirectory = @Suppress("UnstableApiUsage") project.layout.settingsDirectory.dir("build"),
       buildEventsListenerRegistry = buildEventsListenerRegistry,
     )
@@ -480,4 +481,11 @@ public abstract class MavenPublishBaseExtension @Inject constructor(
     val extras = checkNotNull(project.extensions.findByType(ExtraPropertiesExtension::class.java))
     return if (extras.has(propertyName)) extras.get(propertyName)?.toString() else null
   }
+
+  // Resolves a credential value from a Gradle property, falling back to a value set on the
+  // extra properties of the project (for example set by extra["mavenCentralUsername"] = ...).
+  private fun credentialProperty(propertyName: String): Provider<String> =
+    project.providers.gradleProperty(propertyName).orElse(
+      project.provider { findOptionalProperty(propertyName) },
+    )
 }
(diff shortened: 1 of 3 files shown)


plugin/src/main/kotlin/com/vanniktech/maven/publish/MavenPublishBaseExtension.kt
```diff
@@ -120,8 +121,8 @@ public abstract class MavenPublishBaseExtension @Inject constructor(
     }
 
     val buildService = project.registerMavenCentralBuildService(
-      repositoryUsername = project.providers.gradleProperty("mavenCentralUsername"),
-      repositoryPassword = project.providers.gradleProperty("mavenCentralPassword"),
+      repositoryUsername = credentialProperty("mavenCentralUsername"),
+      repositoryPassword = credentialProperty("mavenCentralPassword"),
```
Goooler, line 125:
```suggestion
      repositoryUsername = findOptionalProperty("mavenCentralUsername"),
      repositoryPassword = findOptionalProperty("mavenCentralPassword"),
```

We can call `findOptionalProperty` directly.

plugin/src/main/kotlin/com/vanniktech/maven/publish/MavenPublishBaseExtension.kt
```diff
@@ -480,4 +481,11 @@ public abstract class MavenPublishBaseExtension @Inject constructor(
     val extras = checkNotNull(project.extensions.findByType(ExtraPropertiesExtension::class.java))
     return if (extras.has(propertyName)) extras.get(propertyName)?.toString() else null
   }
+
+  // Resolves a credential value from a Gradle property, falling back to a value set on the
+  // extra properties of the project (for example set by extra["mavenCentralUsername"] = ...).
+  private fun credentialProperty(propertyName: String): Provider<String> =
+    project.providers.gradleProperty(propertyName).orElse(
+      project.provider { findOptionalProperty(propertyName) },
+    )
```
Goooler:
```suggestion
```

Goooler (commented):
It would be great to add tests to cover these cases.

ingokegel:
Thanks for the review!

I'll add test coverage for credential resolution from gradle.properties, -P, ORG_GRADLE_PROJECT_* and extra["mavenCentralUsername"], including that the Gradle property wins when both are set.

credentialProperty has to stay though. findOptionalProperty returns String?, but registerMavenCentralBuildService needs a Provider<String>, so it can't be passed directly. And gradleProperty() covers -P and ORG_GRADLE_PROJECT_*,  while ExtraPropertiesExtension doesn't do this. Dropping credentialProperty would break credentials for everyone passing them that way in CI, which is the common case. It also ensures the correct fallback. Calling findOptionalProperty alone would let an extra[...] value override one from gradle.properties.

